PR #8, opened 2023-01-13
Importkey fixes

svolos:
When a key is not passed as CLI attribute
- Use of crypto/rand rather than math/rand to sample a fresh key
- Store the key in the secretKey variable with the correct scope


KenGordon (approved):
Looks good. Checked with Stavros that crypt/rand does not support seeding.

PR #10, opened 2023-02-06
Ken/uvminfo

KenGordon:
Use the environment provided uvm reference info (cosesign1 wrapped json containing the uvm launch measurement) and amd certificates. Update to latest MAA to match.


pkg/attest/maa.go
```diff
@@ -50,33 +60,56 @@ type attestSNPRequestBody struct {
 // and the certificate chain, (ii) the  runtime data (base64 URL encoding of the public
 // wrapping key), (iii) the inittime data (base64 URL encoding of the security policy),
 // and (iv) a nonce
-func newAttestSNPRequestBody(snpAttestationReport, vcekCertChain, policyBlob, keyBlob []byte) (*attestSNPRequestBody, error) {
+func newAttestSNPRequestBody(snpAttestationReport []byte, vcekCertChain []byte, policyBlob []byte, keyBlob []byte, uvmReferenceInfo []byte) (*attestSNPRequestBody, error) {
```
svolos, line 63: Can we just remove the commented out lines?

pkg/attest/maa.go
```diff
@@ -7,6 +7,8 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+
+	// "io/ioutil" needed when generating test data
```
svolos: can we please remove commented out lines? or use a flag to decide whether to print or not

pkg/attest/data/uvm_host_amd_certificate.json
```diff
@@ -0,0 +1,6 @@
+{
```
svolos: Are these test data used by attest_test ? It may make sense to rename data to test_data ?

pkg/attest/certcache.go
```diff
@@ -3,6 +3,13 @@
 
 package attest
 
+/*
```
svolos, line 6: I think we may need to keep this around, as it is the only way to request certificates from the AMD cert cache. But we can leave this for another PR where we revisit this.

pkg/attest/attest.go
```diff
@@ -53,11 +58,26 @@ func Attest(certCache CertCache, maa MAA, inittimeDataBytes []byte, runtimeDataB
...
+		return "", errors.Wrap(err, "decoding policy from Base64 format failed")
+	}
+	logrus.Debugf("   inittimeDataBytes:    %v", inittimeDataBytes)
+
 	SNPReportBytes, err := FetchSNPReport(fetchRealSNPReport, runtimeDataBytes, inittimeDataBytes)
 	if err != nil {
 		return "", errors.Wrapf(err, "fetching snp report failed")
 	}
 
+	/*
```
svolos, line 72: I think that's where we should be using the AMD Cert cache endpoint just in case we need to fetch fresh certificates.

docker/skr/Makefile
```diff
@@ -0,0 +1,83 @@
+#
```
svolos: why is this file part of the PR ?

cmd/skr/main.go
```diff
@@ -244,6 +232,16 @@ func main() {
 	logFile := flag.String("logfile", "", "Logging Target: An optional file name/path. Omit for console output.")
 	port := flag.String("port", "8080", "Port on which to listen")
 
+	// WARNING!!!
+	// If the security policy does not control the areguments to this process then
```
svolos: areguments -> arguments

svolos (commented):
I have talked with Ken offline and agreed with the changes to the APIs to pass the uvm information and dropping the certcache info. Only few points to consider:
- Change the data directory to test_data 
- Better way to produce the test_data as opposed to commented out lines
- Consider utilizing a default certcache endpoint to fetch fresh certificates. [This can be done via a new PR]

docker/skr/Makefile
KenGordon: Removed

pkg/attest/data/uvm_host_amd_certificate.json
KenGordon: done

cmd/skr/main.go
KenGordon: done

pkg/attest/maa.go
KenGordon: Now in common there is a flag "GenerateTestData" that controls this.

pkg/attest/certcache.go
KenGordon: I have updated the comment.

svolos (approved):

PR #12, opened 2023-02-13
Support for managing secrets with AKV key vaults

svolos:
- MHSM class renamed to AKV class.
- The importkey and encfs/skr sidecar containers now support import and release of RSA-HSM keys. 
- For encfs, the encryption key can be an octet key released from managed HSM or a key derivation from RSA key material released from key vaults using public salt and label.
- The key derivation information is passed via a new object KeyDerivationBlob which may be passed to the import key request or the azure filesystem configuration blobs.


examples/skr/aci-arm-template.json
```diff
@@ -33,10 +33,10 @@
               "environmentVariables": [
                 {
                   "name": "SkrSideCarArgs",
-                  "value": "ewogICAiY2VydGNhY2hlIjogewogICAgICAiZW5kcG9pbnQiOiAiYW1lcmljYXMuYWNjY2FjaGUuYXp1cmUubmV0IiwKICAgICAgInRlZV90eXBlIjogIlNldlNucFZNIiwKICAgICAgImFwaV92ZXJzaW9uIjogImFwaS12ZXJzaW9uPTIwMjAtMTAtMTUtcHJldmlldyIKICAgfSAgICAgIAp9"
```
kapilvgit, line 36: what is this change?

tools/importkey/README.MD
```diff
@@ -1,22 +1,27 @@
...
-When importing a key to the MHSM, a release policy is coupled with the key. The tool follows the schema of AKV MHSM SKR to add a series of claims that need to be satisfied. The authority (MAA endpoint) is also included in the policy so that the MHSM is authorized to release secrets upon receiving valid MAA tokens generated by the MAA endpoint.
+When importing a key to the key vault, a release policy is coupled with the key. The tool follows the schema of AKV SKR API to add a series of claims that need to be satisfied. The authority (MAA endpoint) is also included in the policy so that the vault is authorized to release secrets upon receiving valid MAA tokens generated by the MAA endpoint. 
 
 ## Example import key configuration file
 
 ```json
 {
+    "key_derivation":{
+        "salt": "92a631483ca875aad7e2477da755d58cac3876b77d10bcdd7b33bfa11e7d8b8e",
+        "label": "Model Encryption Key"
```
kapilvgit: Not sure if it matters but the labels are different in key import config and the encrypted file system config.

tools/importkey/main.go
```diff
@@ -94,15 +119,168 @@ func main() {
...
-		KTY:     "oct-HSM",
-		KeyOps:  []string{"encrypt", "decrypt"},
-		K:       base64.RawURLEncoding.EncodeToString(secretKey),
-		KeySize: len(secretKey) * 8,
+	var key interface{}
+	var octKey []byte
+
+	if importKeyCfg.Key.KTY == "RSA-HSM" {
+		var jwKey jwk.RSAPrivateKey
+		if keyRSAPEMFile == "" {
```
kapilvgit, line 127: So is the idea that the user can use this tool either to generate a new RSA key and output a derived key to use for encryption, or provide an RSA key that they have generated outside for importing into AKV?

cmd/skr/main.go
```diff
@@ -184,17 +186,54 @@ func postKeyRelease(c *gin.Context) {
 	skrKeyBlob := skr.KeyBlob{
 		KID:       newKeyReleaseData.KID,
 		Authority: maa,
-		MHSM:      mhsm,
+		AKV:       akv,
 	}
 
-	keyBytes, err := skr.SecureKeyRelease(Identity, skrKeyBlob, EncodedUvmInformation)
+	keyBytes, kty, err := skr.SecureKeyRelease(Identity, skrKeyBlob, EncodedUvmInformation)
```
kapilvgit: Was the key type always returned? Or is this part of a new API?

examples/skr/aci-arm-template.json
svolos: The sidecar used to take the cert cache endpoint information. We now pass an empty string. I guess we could just drop the env. variable alltogether

cmd/skr/main.go
svolos: no, this is part of a new API. I could return a struct that contains both kty and the bytes.

tools/importkey/main.go
svolos, line 127: Yes the tool provides means to randomly generate a key (RSA or OCT) or take as input an RSA key (encoded in ASN.1 DER wrapped in PKCS8 format) or an octet key in hexstring format. When using the -out flag, the tool outputs a symmetric key that can be used for encrypting a filesystem. For oct keys, it is the bytes representation of the hexstring. For RSA keys, it is a key derived from the RSA secret exponent using salt and label provided as input to the import key request.

examples/skr/aci-arm-template.json
kapilvgit: Yes let's drop.

tools/importkey/README.MD
svolos: i am making them consistent throughout . Thanks for the catch

cmd/skr/main.go
svolos: I now return a JWK key as this SecureKeyRelease API may be used by external appications if skr is used as library. So it makes sense to use JWK as the format of the response.

kapilvgit (approved):

examples/skr/aci-arm-template.json
```diff
@@ -116,7 +110,7 @@
                   "value": "eyJrZXlzIjpbeyJlIjoiQVFBQiIsImtleV9vcHMiOlsiZW5jcnlwdCJdLCJraWQiOiJOdmhmdXEyY0NJT0FCOFhSNFhpOVByME5QXzlDZU16V1FHdFdfSEFMel93Iiwia3R5IjoiUlNBIiwibiI6InY5NjVTUm15cDh6Ykc1ZU5GdURDbW1pU2VhSHB1akcyYkNfa2VMU3V6dkRNTE8xV3lyVUp2ZWFhNWJ6TW9PMHBBNDZwWGttYnFIaXNvelZ6cGlORExDbzZkM3o0VHJHTWVGUGYyQVBJTXUtUlNyek41NnF2SFZ5SXI1Y2FXZkhXay1GTVJEd0FlZnlOWVJIa2RZWWtnbUZLNDRoaFVkdGxDQUtFdjVVUXBGWmp2aDRpSTlqVkJkR1lNeUJhS1FMaGpJNVdJaC1RRzZaYTVzU3VPQ0ZNbm11eXV2TjVEZmxwTEZ6NTk1U3MtRW9CSVktTmlsNmxDdHZjR2dSLUlialVZSEFPczVhamFtVHpnZU84a3gzVkNFOUhjeUtteVVac2l5aUY2SURScDJCcHkzTkhUakl6N3Rta3BUSHg3dEhuUnRsZkUyRlV2MEI2aV9RWWxfWkE1USJ9XX0="
                 }
               ],
-              "image": "<registry-name>/aci-skr:1.1",
+              "image": "<registry-name>/skr:latest",
```
KenGordon: Can we not use latest as a tag? It encourages bad practice.

TECH-INFO.md
```diff
@@ -10,15 +10,20 @@ This tool takes an argument such as this:
...
+            "azure_url_private": true,
             "key": {
                  "kid": "EncryptedFilesystemsContainer",
+                 "kty": "RSA-HSM",
                  "authority": {
                      "endpoint": "sharedneu.neu.attest.azure.net"
                  },
-                 "mhsm": { 
-                     "endpoint": "anhsmname.managedhsm.azure.net"
+                 "akv": {
```
KenGordon, line 20: Does this change anything currently exposed to anyone?

svolos: this PR has made changes to the APIs including configuration schemas. So, yes, deployment of new container groups using the new versions of sidecars will have to adapt to the new schema for the sidecar args and SecureKeyRelease APIs.

KenGordon (approved):

PR #22, opened 2023-03-17
Automatically use /dev/sev vs /dev/sev-guest and appropriate IOCTLs t…

KenGordon:
Automatically use /dev/sev vs /dev/sev-guest and appropriate IOCTLs to support transition from 5.15 to 6.* kernels.

5.15 and 6.1 kernels have quite different ways to get an attestation report. This version of get-snp-report notices which device is exposed and then uses the appropriate IOCTLs such that the same container image can support both.


tools/get-snp-report/get-snp-report6.c
```diff
@@ -0,0 +1,73 @@
...
+
+    fd = open("/dev/sev-guest", O_RDWR | O_CLOEXEC);
+
+    if (fd < 0) {
+        fprintf(stdout, "Failed to open /dev/sev-guest\n");        
+        return false;
+    }
+
+	// this is the request, mostly the report data, vmpl
+	snp_report_req snp_request;
```
svolos: can you please ensure that tabspace is consistent ? There are bunch of lines that seem to have more white space.

svolos (changes requested):
Please ensure that there is consistent amount of white tabbing space.

tools/get-snp-report/get-snp-report6.c
KenGordon: I have removed all the tabs from the get-snp-report*.c and the snp-ioctl*.h

svolos (approved):

PR #23, opened 2023-04-06
Support file rather than env based scheme to acquire policy, uvm info and certs.

KenGordon:
Support file rather than env based scheme to acquire policy, uvm info and certs.

To match https://github.com/microsoft/hcsshim/pull/1708

This PR supports both methods to decouple testing/deployment. There will be a subsequent PR to remove support for the environment variable scheme.

This PR also includes a debug version of the skr container that allows ssh-ing into the containers for debugging purposes.

### DIFF
--- a/pkg/common/info.go
+++ b/pkg/common/info.go
@@ -156,12 +156,12 @@ func GetUvmInfomationFromFiles() (UvmInformation, error) {
 
 	encodedUvmInformation.EncodedSecurityPolicy, err = readSecurityContextFile(securityContextDir, PolicyFilename)
 	if err != nil {
-		return encodedUvmInformation, errors.Wrapf(err, "reading host amd cert failed")
+		return encodedUvmInformation, errors.Wrapf(err, "reading security policy failed")
 	}
 
 	encodedUvmInformation.EncodedUvmReferenceInfo, err = readSecurityContextFile(securityContextDir, ReferenceInfoFilename)
 	if err != nil {
-		return encodedUvmInformation, errors.Wrapf(err, "reading host amd cert failed")
+		return encodedUvmInformation, errors.Wrapf(err, "reading uvm reference info failed")
 	}
 
 	if GenerateTestData {
--- a/pkg/common/info.go
+++ b/pkg/common/info.go
@@ -9,6 +9,7 @@ import (
 
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


pkg/common/info.go
```diff
@@ -90,3 +110,64 @@ func GetUvmInfomation() (UvmInformation, error) {
...
+	ReferenceInfoFilename      = "reference-info-base64"
+)
+
+func readSecurityContextFile(dir string, filename string) (string, error) {
+	targetFilename := filepath.Join(dir, filename)
+	blob, err := os.ReadFile(targetFilename)
+	if err != nil {
+		return "", err
+	}
+	return string(blob), nil
```
BryceDFisher, line 129: Unless you need to guarantee that a short read is never returned, this could be simplified to:
`blob, err := os.ReadFile(targetFilename)
return string(blob), err`

pkg/common/info.go
```diff
@@ -90,3 +110,64 @@ func GetUvmInfomation() (UvmInformation, error) {
...
+		certChain, err := THIMtoPEM(encodedHostCertsFromTHIM)
+		if err != nil {
+			return encodedUvmInformation, err
+		}
+		encodedUvmInformation.CertChain = certChain
+	}
+
+	encodedUvmInformation.EncodedSecurityPolicy, err = readSecurityContextFile(securityContextDir, PolicyFilename)
+	if err != nil {
+		return encodedUvmInformation, errors.Wrapf(err, "reading host amd cert failed")
```
BryceDFisher: I think this was a copy/paste.  Should be an error about reading the policy

pkg/common/info.go
```diff
@@ -90,3 +110,64 @@ func GetUvmInfomation() (UvmInformation, error) {
...
+	}
+
+	encodedUvmInformation.EncodedUvmReferenceInfo, err = readSecurityContextFile(securityContextDir, ReferenceInfoFilename)
+	if err != nil {
+		return encodedUvmInformation, errors.Wrapf(err, "reading host amd cert failed")
```
BryceDFisher: I think this was a copy/paste. Should be an error about reading the reference info

hgarvison (approved):

tools/securitypolicydigest/main.go
```diff
@@ -11,30 +11,61 @@ import (
...
 	// variables declaration
 	var encodedSecurityPolicy string
+	var verbose bool
+	var filename string
 
 	// flags declaration using flag package
 	flag.StringVar(&encodedSecurityPolicy, "p", "", "Security policy in base64-encoded string format")
-	flag.Parse() // after declaring flags we need to call it
-
+	flag.StringVar(&filename, "f", "", "file containing ecurity policy in base64-encoded string format")
```
hgarvison: *security policy

tools/securitypolicydigest/main.go
```diff
@@ -11,30 +11,61 @@ import (
...
 	inittimeDataBytes, err := base64.StdEncoding.DecodeString(encodedSecurityPolicy)
 	if err != nil {
-		fmt.Println("Could not decode ")
+		fmt.Println("Could not decode - required base64 (Std, with padding) encoding")
 		os.Exit(1)
 	}
 
-	fmt.Printf("inittimeData %s", string(inittimeDataBytes))
+	if verbose {
+		fmt.Printf("inittimeData %s", string(inittimeDataBytes))
```
hgarvison: End with a newline so it doesn't run into the next line when verbose

tools/securitypolicydigest/main.go
```diff
@@ -11,30 +11,61 @@ import (
...
+			fmt.Printf("Could not decode file %s\n", filename)
+			os.Exit(1)
+		}
+		encodedSecurityPolicy = string(encodedSecurityPolicyBytes)
+	}
+
 	inittimeDataBytes, err := base64.StdEncoding.DecodeString(encodedSecurityPolicy)
 	if err != nil {
-		fmt.Println("Could not decode ")
+		fmt.Println("Could not decode - required base64 (Std, with padding) encoding")
```
hgarvison: A nit, but I would say "Could not decode: base64 encoding (Std, with padding) required"

svolos (approved):
This PR's functionality has been tested by running the skr example on clusters that pass the uvm information via files or env. variables.

tools/securitypolicydigest/main.go
KenGordon: Done

tools/securitypolicydigest/main.go
KenGordon: done

tools/securitypolicydigest/main.go
KenGordon: done

hgarvison (approved):

PR #24, opened 2023-04-06
Updates to check local certs, fetch remote certs, and compare TCB Versions

hgarvison:


KenGordon:
Can you not check that this field https://github.com/microsoft/confidential-sidecar-containers/blob/c85bfab6cdea1245e7a6ce84eeec81386b5c3fe5/pkg/common/info.go#L36 as discussed here https://learn.microsoft.com/en-us/azure/security/fundamentals/trusted-hardware-identity-management rather than having to parse the certs? 

Given you start with a known version, then must fetch a given version at all times you can maintain the tcb you have without parsing the cert.

Note also that the remote cache is not the same as the LO THIM cache, thus we need to know the type of endpoint. This model is in place in the previous CertCache which I would like you to extend for this purpose. Have it able to go to an L0 proxy, so move that THIMtoPEM function into the cache code and enable a new type THIM that expects that format. Keep the code that can handle both AMD and Azure endpoints (ir this you are calling THIM but is the non box local but azure local cache. Please note that the L0 and that cache are different.

KenGordon:
Stavros’s CertCache that you should extend https://github.com/microsoft/confidential-sidecar-containers/blob/84d73d2d2daf910e92cb8d3fe7d9b028aa48e7c6/pkg/attest/certcache.go

To match please reinstate

https://github.com/microsoft/confidential-sidecar-containers/blob/84d73d2d2daf910e92cb8d3fe7d9b028aa48e7c6/cmd/skr/main.go#L30

To let clients specify the fallback end point. Note that that includes the type of the endpoint as they all work differently.

KenGordon:
If we were being very clever we’d either allow the client to specify several fallbacks, or fallback to the (initially unavailable) L1 proxy to the L0 THIM.

cmd/remotefs/azurefs.go
```diff
@@ -355,7 +355,7 @@ func MountAzureFilesystems(tempDir string, info RemoteFilesystemsInformation) (e
 	Identity = info.AzureInfo.Identity
 
 	// Retrieve the incoming encoded security policy, cert and uvm endorsement
-	EncodedUvmInformation, err = common.GetUvmInfomation()
+	EncodedUvmInformation, err = common.GetUvmInformation()
```
KenGordon: Oops.

pkg/common/info.go
```diff
@@ -26,9 +32,122 @@ type UvmInformation struct {
...
+	}
+	return nil, errors.New("No UVM certificate chain found")
+}
+
+// parses the VCEK certificate to return the TCB version
+// fields reprocessed back into a uint64 to compare against the `ReportedTCB` in the fetched attestation report
+func (u *UvmInformation) ParseVCEK() ( /*tcbVersion*/ uint64, error) {
+	vcekCert, err := u.GetParsedVCEK()
+	if err != nil {
+		return 0, err
```
KenGordon: I don’t think we need/should do this. Instead remember the TCB of the certs you have - THIM tells you, all the other sources you have to specify when requesting them.

pkg/common/info.go
```diff
@@ -65,11 +184,14 @@ func THIMtoPEM(encodedHostCertsFromTHIM string) (string, error) {
 	return certsString, nil
 }
 
-func GetUvmInfomation() (UvmInformation, error) {
+func GetUvmInformation() (UvmInformation, error) {
 	var encodedUvmInformation UvmInformation
+	encodedUvmInformation.RemoteTHIMServiceURL = os.Getenv("UVM_THIM_SERVICE_URL") + "/vcek/v1/"
```
KenGordon: This is user config, not a system thing. It needs to be parsed out of the json “AzureInformation” as per https://github.com/microsoft/confidential-sidecar-containers/blob/c85bfab6cdea1245e7a6ce84eeec81386b5c3fe5/cmd/skr/main.go#L289

pkg/common/info.go
```diff
@@ -65,11 +184,14 @@ func THIMtoPEM(encodedHostCertsFromTHIM string) (string, error) {
 	return certsString, nil
 }
 
-func GetUvmInfomation() (UvmInformation, error) {
+func GetUvmInformation() (UvmInformation, error) {
```
KenGordon: The ContainerPlat people would insist fixing my typo go in a discrete PR. :)

BryceDFisher: An API purist would insist that this is part of the external (to the package) API and thus can never be fixed (laugh)

pkg/common/info.go
```diff
@@ -26,9 +32,122 @@ type UvmInformation struct {
...
+		}
+		// snpSPL
+		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.3.3") {
+			snp = fmt.Sprintf("snpSPL=%d&", ext.Value[2])
+			tcbValues[6] = ext.Value[2]
+		}
+		// ucodeSPL
+		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.3.8") {
+			ucode = fmt.Sprintf("ucodeSPL=%d&", ext.Value[2])
+			tcbValues[7] = ext.Value[2]
```
BryceDFisher: If this stays, the index values need to be turned into constant vars so that they can be reused and future devs can know what is going where.

pkg/common/info.go
```diff
@@ -26,9 +32,122 @@ type UvmInformation struct {
...
+		return 0, err
+	}
+
+	tcbValues := make([]byte, 8) // TCB version is 8 bytes
+	var hwid, ucode, bl, tee, snp string
+	// parse extensions to update the THIM URL and get TCB Version
+	for _, ext := range vcekCert.Extensions {
+		// productName
+		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.2") {
+			u.RemoteTHIMServiceURL += string(ext.Value) + "/"
```
BryceDFisher: This function will only work once.  If it is ever executed a second time, this url will include the host, path, and query string twice.  I recommend building the URL using the url.URL object in the `UvmInformation` type and then set the Path and RawQuery attributes in this function.  When needed, call url.String() to get the final version for querying.

pkg/attest/attest_test.go
```diff
@@ -567,3 +568,93 @@ func Test_MAA(t *testing.T) {
...
...9eNxhIySb2npJfgnqz55I0u33wh4r0ZNQ\neTGfw03MBUtyuzGesGkcw+loqMaq1qR4tjGbPYxCvpCq7+OgpCCoMNit2uLo9M18\nfHz10lOMT8nWAUvRZFzteXCm+7PHdYPlmQwUw3LvenJ/ILXoQPHfbkH0CyPfhl1j\nWhJFZasCAwEAAaN+MHwwDgYDVR0PAQH/BAQDAgEGMB0GA1UdDgQWBBSFrBrRQ/fI\nrFXUxR1BSKvVeErUUzAPBgNVHRMBAf8EBTADAQH/MDoGA1UdHwQzMDEwL6AtoCuG\nKWh0dHBzOi8va2RzaW50Zi5hbWQuY29tL3ZjZWsvdjEvTWlsYW4vY3JsMEYGCSqG\nSIb3DQEBCjA5oA8wDQYJYIZIAWUDBAICBQChHDAaBgkqhkiG9w0BAQgwDQYJYIZI\nAWUDBAICBQCiAwIBMKMDAgEBA4ICAQC6m0kDp6zv4Ojfgy+zleehsx6ol0ocgVel\nETobpx+EuCsqVFRPK1jZ1sp/lyd9+0fQ0r66n7kagRk4Ca39g66WGTJMeJdqYriw\nSTjjDCKVPSesWXYPVAyDhmP5n2v+BYipZWhpvqpaiO+EGK5IBP+578QeW/sSokrK\ndHaLAxG2LhZxj9aF73fqC7OAJZ5aPonw4RE299FVarh1Tx2eT3wSgkDgutCTB1Yq\nzT5DuwvAe+co2CIVIzMDamYuSFjPN0BCgojl7V+bTou7dMsqIu/TW/rPCX9/EUcp\nKGKqPQ3P+N9r1hjEFY1plBg93t53OOo49GNI+V1zvXPLI6xIFVsh+mto2RtgEX/e\npmMKTNN6psW88qg7c1hTWtN6MbRuQ0vm+O+/2tKBF2h8THb94OvvHHoFDpbCELlq\nHnIYhxy0YKXGyaW1NjfULxrrmxVW4wcn5E8GddmvNa6yYm8scJagEi13mhGu4Jqh\n3QU3sf8iUSUr09xQDwHtOQUVIqx4maBZPBtSMf+qUDtjXSSq8lfWcd8bLr9mdsUn\nJZJ0+tuPMKmBnSH860llKk+VpVQsgqbzDIvOLvD6W1Umq25boxCYJ+TuBoa4s+HH\nCViAvgT9kf/rBq1d+ivj6skkHxuzcxbk1xv6ZGxrteJxVH7KlX7YRdZ6eARKwLe4\nAFZEAwoKCQ==\n-----END CERTIFICATE-----\n",
```
BryceDFisher: If its easier, you can use the "`" character to do multiline strings.

pkg/common/info.go
```diff
@@ -26,9 +32,122 @@ type UvmInformation struct {
...
+		}
+		// HwID
+		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.4") {
+			hwid = hex.EncodeToString(ext.Value) + "?"
+		}
+		// blSPL
+		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.3.1") {
+			bl = fmt.Sprintf("blSPL=%d", ext.Value[2])
```
BryceDFisher: 2nd byte of the pkix extension? This is throwing me off.  Could use a comment in here as to why.

pkg/common/info.go
hgarvison: I actually started it using the url package, in particular the parameters list didn't format correctly to match the url that I saw in the test code, so I switch it to this method. I can attempt it again.

pkg/attest/attest_test.go
hgarvison: I just copied this cert chain from already existing test code, which had the string in one line.

pkg/common/info.go
hgarvison: I was told that there would be an env var (in the future?), which is why this was created, but if this should be parsed out of the AzureInformation, does that mean a new field is being added to that struct, if so, is there a name for the json field to match that I should know about?

pkg/common/info.go
hgarvison: @BryceDFisher can you look at the comment I added to the bottom of this function and tell me if that matters or if there's a workaround for that weird string formatting?

pkg/common/info.go
KenGordon: These sidecars receive config as base64 encoded json structures (AzureInformation for example) and from the UVM by some means (my PR changes that from env variables to files). You should be reinstating the field in AzureInformation that describes how to find the certs. You should not add that to UvmInformation as it is config about the outside world, provided to the container by “not the UVM”. Yes, today the certs are there as they came from the UVM, but your work is to add an EXTRA source of certs.

pkg/common/info.go
```diff
@@ -94,55 +109,65 @@ func (u *UvmInformation) ParseVCEK() ( /*tcbVersion*/ uint64, error) {
...
-	var hwid, ucode, bl, tee, snp string
+	v := url.Values{}
+	var hwid string //, ucode, bl, tee, snp string
 	// parse extensions to update the THIM URL and get TCB Version
 	for _, ext := range vcekCert.Extensions {
+		fmt.Println(ext.Value)
 		// productName
 		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.2") {
-			u.RemoteTHIMServiceURL += string(ext.Value) + "/"
+			u.RemoteTHIMServiceURL.Path += string(ext.Value) + "/"
```
BryceDFisher: This has the same problem as before.  If this function get's called twice, the value of `u.RemoteTHIMServiceURL.Path` will get appended to twice.

hgarvison: This is out of date now

pkg/common/info.go
```diff
@@ -94,55 +109,65 @@ func (u *UvmInformation) ParseVCEK() ( /*tcbVersion*/ uint64, error) {
...
 		// snpSPL
 		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.3.3") {
-			snp = fmt.Sprintf("snpSPL=%d&", ext.Value[2])
-			tcbValues[6] = ext.Value[2]
+			//snp = fmt.Sprintf("snpSPL=%d&", ext.Value[2])
+			v.Set("snpSPL", fmt.Sprintf("%d", ext.Value[x509CertExtensionsValuePos]))
+			tcbValues[TcbSnpSpl] = ext.Value[x509CertExtensionsValuePos]
 		}
 		// ucodeSPL
 		if strings.Contains(ext.Id.String(), "1.3.6.1.4.1.3704.1.3.8") {
```
BryceDFisher: Just noticed, these can all be `} else if {...` instead of checking the value each time.  You can also switch to a 

```go
switch ext.Id.String() {
case "...":
case "1.3.6.1.4.1.3704.1.3.3":
case "1.3.6.1.4.1.3704.1.3.8":
}
```

instead to only evaluate the `.String()` function once.

cmd/skr/main.go
```diff
@@ -221,7 +226,7 @@ func setupServer(identity common.Identity) *gin.Engine {
 	// the certificate chain endording the signing key of the hardware attestation.
 	// Hence, these APIs are exposed only if the platform certificate information
 	// has been provided at startup time.
-	if EncodedUvmInformation.CertChain != "" {
+	if ServerCertCache.Endpoint != "" {
```
BryceDFisher: We want to include the MAA endpoint if a URL is configured or if the CertChain is present via other means (like the file mounted from the UVM)

pkg/attest/attest.go
```diff
@@ -90,7 +83,23 @@ func Attest(maa MAA, runtimeDataBytes []byte, uvmInformation common.UvmInformati
 		return "", errors.Wrapf(err, "failed to deserialize attestation report")
 	}
 
-	vcekCertChain := []byte(uvmInformation.CertChain)
+	// At this point check that the TCB of the cert chain matches that reported so we fail early or
+	// fetch fresh certs by other means.
+	var vcekCertChain []byte
+	thimTcbm, err := strconv.ParseUint(uvmInformation.ThimCerts.Tcbm, 10, 64)
+	if err != nil {
+		return "", errors.Wrap(err, "converting TCBM string to uint64 failed")
```
BryceDFisher: "Unable to convert TCBM from THIM certificates to a uint64"?

pkg/attest/certcache.go
```diff
@@ -63,29 +72,56 @@ func (certCache CertCache) retrieveCertChain(chipID string, reportedTCB uint64)
...
 		var fullCertChain []byte
 		fullCertChain = append(fullCertChain, vcekPEMBytes[:]...)
 		fullCertChain = append(fullCertChain, certChainPEMBytes[:]...)
 
-		return fullCertChain, nil
+		return fullCertChain, thimCerts, nil
+	} else if certCache.THIM {
+		// AMD THIM cert cache endpoint returns THIM Certs object
+		// Is TEEType the same as productName in this case or should I add a field for that?
+		uri = fmt.Sprintf(AmdTHIMRequestURITemplate, certCache.Endpoint, certCache.TEEType, chipID, reportedTCBBytes[7], reportedTCBBytes[6], reportedTCBBytes[1], reportedTCBBytes[0])
```
BryceDFisher: It would still be a good idea to make the indexes constants to aid future development.

pkg/attest/certcache.go
```diff
@@ -63,29 +72,56 @@ func (certCache CertCache) retrieveCertChain(chipID string, reportedTCB uint64)
...
+
+		thimCerts, err = common.THIMtoPEM(string(THIMCertsBytes))
+		if err != nil {
+			return nil, thimCerts, err
+		}
+
+		// constuct full chain by appending the VCEK cert to the cert chain
+		var fullCertChain []byte
+		fullCertChain = append(fullCertChain, []byte(thimCerts.VcekCert)[:]...)
+		fullCertChain = append(fullCertChain, []byte(thimCerts.CertificateChain)[:]...)
```
BryceDFisher: Probably not a huge issue here, but you can avoid additional memory allocations by either directly allocating the memory up front.

Most memory efficient but does a copy twice:

```go
fullCertChain := make([]byte,0,len(thimCerts.VcekCert)+len(thimCerts.CertificateChain))
```

or make the `fullCertChain` in 1 step, but over allocates memory:

```go
fullCertChain := append([]byte(thimCerts.VcekCert), []byte(thimCerts.CertificateChain)...)
```

BryceDFisher: https://go.dev/play/p/agKqXGHYHlH

cmd/skr/main.go
BryceDFisher:
```go
if serverCertCache.Endpoint != "" || encodedUvmInformation.CertChain != "" {
```

BryceDFisher (approved):

pkg/common/info.go
```diff
@@ -20,65 +20,64 @@ import (
...
 	}
 
-	var certsFromTHIM THIMCerts
 	err = json.Unmarshal(hostCertsFromTHIM, &certsFromTHIM)
 	if err != nil {
-		return "", errors.Wrapf(err, "json unmarshal platform certs failed")
+		return certsFromTHIM, errors.Wrapf(err, "json unmarshal platform certs failed")
 	}
 
-	certsString := certsFromTHIM.VcekCert + certsFromTHIM.CertificateChain
```
KenGordon, line 58: This line, which you have moved out to the caller someplace. Is the main point of this function

pkg/common/info.go
```diff
@@ -20,65 +20,64 @@ import (
...
-
 type THIMCerts struct {
 	VcekCert         string `json:"vcekCert"`
 	Tcbm             string `json:"tcbm"`
 	CertificateChain string `json:"certificateChain"`
 	CacheControl     string `json:"cacheControl"`
 }
 
-func THIMtoPEM(encodedHostCertsFromTHIM string) (string, error) {
+func THIMtoPEM(encodedHostCertsFromTHIM string) (THIMCerts, error) {
```
KenGordon: This function is supposed to return a PEM, hence the name.

pkg/attest/certcache.go
```diff
@@ -63,29 +79,52 @@ func (certCache CertCache) retrieveCertChain(chipID string, reportedTCB uint64)
...
 
 		// constuct full chain by appending the VCEK cert to the cert chain
-		var fullCertChain []byte
-		fullCertChain = append(fullCertChain, vcekPEMBytes[:]...)
-		fullCertChain = append(fullCertChain, certChainPEMBytes[:]...)
+		fullCertChain := append(vcekPEMBytes, certChainPEMBytes[:]...)
 
-		return fullCertChain, nil
+		return fullCertChain, thimCerts, nil
+	} else if certCache.THIM {
```
KenGordon: Is this the L0/L1 THIM or an Azure Cache? If it is the L0/L1 THIM then it doesn't match the request form here https://dev.azure.com/msazure/ContainerPlatform/_git/containerd.cri?path=/pkg/containerd/opts/spec.go&version=GBmain&line=31&lineEnd=32&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents and if it is an Azure cache then the code is surely already in the else case below (was line 80, now line 115)

hgarvison: I was informed that the THIM endpoint would match the format in the AMD documentation https://www.amd.com/system/files/TechDocs/57230.pdf hence the extra conditional, if that's not the case, I can update the the request form to match that format.

KenGordon: That is the AMD endpoints. The local THIM ones are not the same. The certs ect are as described but not the json. Try this sort of request

curl GET "http://169.254.169.254/metadata/THIM/amd/certification" -H "Metadata: true” and compare it with replies from the AMD one.

KenGordon: (Note that the AMD endpoints are case sensitive)

KenGordon: Basically there are four sources and three formats for the cert infomation:
 - AMD servers (probably throttled) - you need to ask by TCB
 - Azure Cache - you need to ask by TCB
 - Local THIM endpoint - returns current (at time of asking - there is still a race) TCB
 - UVM/hcsshim provided THIM certs - returns TCB of container startup time. These are the same certs at the Local endpoint, just fetched by hcsshim.

hgarvison: Based on this, are you recommending that I update the retrieveCertChain function to get the TCB from the local THIM endpoint and then parse/use that value to get the certChain (as it was doing before my edits)? Or are you suggesting that I make a separate function to get the tcb value from the local THIM endpoint and call that whenever we need to update/compare the TCB value with the one in the SNP report?

KenGordon: You should only ever obtain the TCB from an attestation report. I am just telling you the properties of the THIM ones. Given you have a TCB from a report and it does not match you need to fetch a fresh set of certs, either from the live local THIM (and you will need to check the TCB matches and maybe get a new report if it does not) or from a remote one (From AzCache or AMD) using the TCB in the report to specify what one you need. The AzCache and AMD versions of that fetching are in Stavros's original CertCache code.

KenGordon: For the local and fetched by hcsshim certs you have two races. One between obtaining the certs and getting the report and another between getting the report and obtaining the local certs. You should probably always get the certs first and retry BOTH the cert fetch and the report fetch if the certs do not match the report.

hgarvison: So I think I've addressed the snp/cache tcb checks, but I still need clarification on when to use the local THIM endpoint to refresh the cert cache. Thanks!

KenGordon: When the AzureInformation that provides the endpoints provides one of the (local) THIM type. That will happen when the ACI/SF team add such an endpoint and users choose it.

hgarvison: What I meant was, how is the application supposed to know if the endpoint provided is a local THIM type? Is the flag I made in CertCache struct acceptable or is there another method that will provide this?

KenGordon: That is what this is for I think 
	TEEType    string `json:"tee_type,omitempty"`

hgarvison: What would be the value of TEEType that would indicate THIM? "THIM"?

KenGordon: I might call it LocalTHIM. Shiv's team call the thing we call the Azure Cache THIM I believe.

examples/skr/importkeyconfig.json
```diff
@@ -3,12 +3,12 @@
         "kid": "<key-vault-key-identifier>",
         "kty": "<key-type-of-key-imported-to-akv>",
         "authority": {
-            "endpoint": "<authority-of-key-imported-to-key-vault>"
+            "endpoint": "<microsoft-azure-attestation-endpoint>"
```
svolos: the authority can be any entity; not just an MAA instance. Please fix to <authority-of-claims-required-for-key-release-from-key-vault>

pkg/attest/attest.go
```diff
@@ -15,6 +16,49 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func GetSNPReport(securityPolicy string, runtimeDataBytes []byte) ([]byte, []byte, error) {
+	// check if sev device exists on the platform; if not fetch fake snp report
+	var fetchRealSNPReport bool
+	if _, err := os.Stat("/dev/sev"); os.IsNotExist(err) {
```
svolos: @KenGordon , i just realized that this line needs to change and check whether any of the /dev/sev or /dev/sev-guest exists.

pkg/attest/attest.go
```diff
@@ -47,50 +91,55 @@ func RawAttest(inittimeDataBytes []byte, runtimeDataBytes []byte) (string, error
...
+			// if no match after refreshing certs and attestation report, fail
+			if SNPReport.ReportedTCB != uvmInformation.ThimTcbm {
+				return "", errors.New("SNP reported TCB value doesn't match Certificate TCB values")
+			}
+		}
+	} else {
+		vcekCertChain = []byte(uvmInformation.CertChain)
+	}
 
 	/* TODO: to support use outside of Azure add code to fetch the AMD certs here */
```
svolos: i guess this comment needs to go away because this PR does exactly what the TODO says.

pkg/attest/certcache.go
```diff
@@ -38,7 +46,7 @@ type CertCache struct {
 // retrieveCertChain interacts with the cert cache service to fetch the cert chain of the
 // chip identified by chipId running firmware identified by reportedTCB. These attributes
 // are retrived from the attestation report.
-func (certCache CertCache) retrieveCertChain(chipID string, reportedTCB uint64) ([]byte, error) {
```
svolos, line 41: What is the purpose of the string return variable? Can you update the retrieveCertChain description to explain what the return values are? []byte is the full cert chain and string is related to thim ?

svolos (changes requested):

pkg/attest/attest.go
KenGordon: Good point.

pkg/attest/certcache.go
```diff
@@ -24,6 +24,14 @@ const (
 	AzureCertCacheRequestURITemplate = "https://%s/%s/certificates/%s/%s?%s"
 	AmdVCEKRequestURITemplate        = "https://%s/%s/%s?ucodeSPL=%d&snpSPL=%d&teeSPL=%d&blSPL=%d"
 	AmdCertChainRequestURITemplate   = "https://%s/%s/cert_chain"
+	LocalTHIMCertEndpoint            = "http://169.254.169.254/metadata/THIM/amd/certification"
```
KenGordon: This needs to come as config, same as the others. Since the L1 version is yet to be defined I'd assume the whole path is must be supplied.

pkg/common/info.go
```diff
@@ -20,28 +21,26 @@ import (
 
 const GenerateTestData = false
 
-// Information supplied by the UVM specific to running Pod
-
 type UvmInformation struct {
 	EncodedSecurityPolicy   string // customer security policy
 	CertChain               string // platform certificates for the actual physical host, ascii PEM
 	EncodedUvmReferenceInfo string // endorsements for the particular UVM image
+	ThimTcbm                uint64 // TCBM for the particular UVM image
```
KenGordon: UVM images do not have a TCBM. That is a property of the report and the certs.

pkg/common/info.go
```diff
@@ -62,23 +61,29 @@ func THIMtoPEM(encodedHostCertsFromTHIM string) (string, error) {
...
 	}
 
 	if encodedHostCertsFromTHIM != "" {
-		certChain, err := THIMtoPEM(encodedHostCertsFromTHIM)
+		certsFromTHIM, tcbmFromTHIM, err := THIMtoPEM(encodedHostCertsFromTHIM)
 		if err != nil {
 			return encodedUvmInformation, err
 		}
-		encodedUvmInformation.CertChain = certChain
+		thimTcbm, err := strconv.ParseUint(tcbmFromTHIM, 10, 64)
```
KenGordon: you might move extracting the tcb into THIMtoPEM and call that something like decodeTHIM. bIn the case you fetch a THIM though it will be plain json, not base64 encoded I believe.

pkg/attest/certcache.go
hgarvison: And where in the config would that be supplied? The encodedHostCertsFromTHIM (demarshalled into the THIMCerts struct)?

Dockerfile.skr
```diff
@@ -0,0 +1,20 @@
+FROM golang:latest as build
```
svolos: Are these changes to the Dockerfile.skr consistent with the version of the build.sh ? What's the reasoning behind this change?

tools/securitypolicydigest/main.go
```diff
@@ -25,7 +25,11 @@ func main() {
 
 	// flags declaration using flag package
 	flag.StringVar(&encodedSecurityPolicy, "p", "", "Security policy in base64-encoded string format")
+<<<<<<< HEAD
+	flag.StringVar(&filename, "f", "", "file containing ecurity policy in base64-encoded string format")
+=======
 	flag.StringVar(&filename, "f", "", "file containing security policy in base64-encoded string format")
```
svolos: merge conflict resolution issue here

.gitignore
```diff
@@ -1,5 +1,9 @@
 **.o
 /azmount
 /remotefs
-/tools/get-snp-report/bin
+
 /bin
+/docker/skr/bin
+/docker/encfs/bin
+/tools/get-snp-report/*
```
KenGordon: There is code there.

tools/get-snp-report/verbose-report.c
```diff
@@ -22,24 +22,31 @@
...
+    {
         report_data_hexstring = argv[1];
     }
 
-    if (supportsDevSev()) {
-        success = fetchAttestationReport5(report_data_hexstring, (void*) &snp_report_hex);
-    } else if (supportsDevSevGuest()) {
-        success = fetchAttestationReport6(report_data_hexstring, (void*) &snp_report_hex);
-    } else {
+    if (supportsDevSev())
```
KenGordon: Please don't. :)

Dockerfile.skr
KenGordon: This doesn't belong here following how Stavros built the repo. It is fine as it is in anycase I believe.

tools/get-snp-report/verbose-report.c
hgarvison: Ah, my vscode auto-formatted this. I can revert it.

pkg/attest/certcache.go
```diff
@@ -63,29 +73,48 @@ func (certCache CertCache) retrieveCertChain(chipID string, reportedTCB uint64)
...
 
 		// constuct full chain by appending the VCEK cert to the cert chain
-		var fullCertChain []byte
-		fullCertChain = append(fullCertChain, vcekPEMBytes[:]...)
-		fullCertChain = append(fullCertChain, certChainPEMBytes[:]...)
+		fullCertChain := append(vcekPEMBytes, certChainPEMBytes[:]...)
 
-		return fullCertChain, nil
+		return fullCertChain, thimTcbm, nil
+	} else if certCache.TEEType == "LocalTHIM" && localThimUri != "" {
```
svolos: I am quite confused regarding the cert cache endpoints and the order at which they are called.

There is the CertChain as part of the uvmInformation which may be stale and hence will require refreshing the cert chain via a cert cache endpoint: local THIM endpoint or a remote cert cache. The remote cert cache may be the Azure Cert Cache or the AMD Cert Cache as denoted by the certCache.AMD flag.

What is the desired priority of the cert cache endpoints? is it local THIM followed by remote cert cache ? or is it remote cert cache followed by local THIM ?  I would expect that the local THIM should be called first.  The current if-then-else statement prioritizes AMD cert cache endpoints over local THIM endpoints, and prioritizes local THIM endpoints over Azure Cert cache endpoints. So, this needs to change based on the desired priority. 

Another issue here is how the code decides whether the local THIM shall be used. Right now, the TEEtype is re-used to pass information whether the LocalTHIM shall be used or not. This is problematic because in the scenario that the remote cache needs to be called, the TEE type will not have the right value (e.g., SevSnpVm.) (For instance, if this  TEEType is set to LocalTHIM when the certcache instance is brought up by the cmd/skr/main.go, then fetching certs from AMD or Azure cert cache will fail due to non-correct TEEType. If the TEEType is set to SevSnpVm, the local thim endpoint will never be called)

The confusion here is made because the CertCache object (initially meant to capture remote cache endpoint information) is being reused for local cert cache endpoint. 

[Less hassle option] One option is to remove local cert cache endpoint functionality from the CertCache:
- I would remove the code block about LocalTHIM calls here and keep the certcache implementation and API as it used to be. 
- The right place to call the local thim shall be in the attest.RefreshCertChain(...., localTHIM = true) . If localTHIM, use the uvmInformation.localThimUri to call using the code 89-102. 
- The attest function shall implement the desired priority by calling attest.RefresthCertChain( ...., true) and checking if the cert chain tcb matches the report's and if not by calling attest.RefresthCertChain( ...., false) which will call the certCache.RetrieveCertChain 

[Preferred option] Another option is to re-visit the CertCache object and have the notion of a LocalTHIM Object (will cotnain the uri), a RemoteCache Object  (will contain all the existing attributes) and a LocalTHIMPriority flag. Then we have LocalTHIMGetCertChain and RemoteCetCertChain methods to remove the implementation of the priority in the current GetCertChain method. The priority (as discussed above) will be implemented by the attest.RefreshCertChain method based on the priority flag specified in the CertCache object.

svolos (changes requested):
How is this new functionality being tested? That means, how can we check the following scenarios: 
(A) If cert chain is stale, the cert chain is retrieved from local THIM
(B) If cert chain is stale, the cert chain is retrieved from local cert cache.

pkg/attest/attest.go
```diff
@@ -15,6 +15,47 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func GetSNPReport(securityPolicy string, runtimeDataBytes []byte) ([]byte, []byte, error) {
+	// check if sev device exists on the platform; if not fetch fake snp report
+	fetchRealSNPReport := true
+	if _, err := os.Stat("/dev/sev"); errors.Is(err, os.ErrNotExist) {
```
KenGordon, line 29: Oh yes, maybe I should do that on main in anycase. Opps.

pkg/attest/certcache.go
hgarvison: I'm okay with this

pkg/attest/certcache.go
```diff
@@ -24,37 +24,50 @@ const (
...
 )
 
 // CertCache contains information about the certificate cache service
 // that provides access to the certificate chain required upon attestation
 type CertCache struct {
-	AMD        bool   `json:"amd,omitempty"`
-	Endpoint   string `json:"endpoint"`
-	TEEType    string `json:"tee_type,omitempty"`
-	APIVersion string `json:"api_version,omitempty"`
+	EndpointType string `json:"endpoint_type,omitempty"` // AMD, AzCache, LocalTHIM
```
KenGordon: I think that the endpoint_type is now required and if the whole thing is missing we should not try to fetch a cert but instead fail with some message about the stale certs. When I ran this container with my last test arm template if fails like this:
root@SandboxHost-638176805756747558:/# cat keyrelease.out 
{"error":"attestation failed: refreshing CertChain failed: HTTP GET failed: Get \"https:////certificates/c5943d9ba2f4837a343b079cc7b65d5669d2967caff818bf4ebc8429fd32d04b782c7cd032971aa0dda2892842d9f59d93a381c0eb47d62d75bb17fdb22cc271/7308000000000003?\": http: no Host in request URL"}root@SandboxHost-638176805756747558:/#

pkg/attest/certcache.go
```diff
@@ -63,29 +76,49 @@ func (certCache CertCache) retrieveCertChain(chipID string, reportedTCB uint64)
...
 
-		return fullCertChain, nil
-	} else {
+		thimCerts, thimTcbm, err := thimCerts.GetLocalCerts(string(THIMCertsBytes))
+		if err != nil {
+			return nil, thimTcbm, err
+		}
+
+		return []byte(thimCerts), thimTcbm, nil
+	default: // case "AzCache": --> to match previous if-else logic and avoid failing when no certcache is supplied
```
KenGordon: As per other comment, if we have not CertCache we have no idea where to send the request. We should not try. The previous logic might have had this problem to, but it always had to have a url as it was how it was done 100% of the time, not just a fallback.

pkg/attest/certcache.go
```diff
@@ -24,37 +24,50 @@ const (
...
+const (
+	BlSplTcbmByteIndex    = 0
+	TeeSplTcbmByteIndex   = 1
+	SnpSplTcbmByteIndex   = 6
+	UcodeSplTcbmByteIndex = 7
 )
 
 // CertCache contains information about the certificate cache service
 // that provides access to the certificate chain required upon attestation
 type CertCache struct {
```
KenGordon: Can we give this (and the file) a new name please? I know this was always called CertCache, but now it is more like a CertFetcher and the thing above caches the result. The previous naming was about the AzCache that this talks to.

svolos (changes requested):
I apologize for not suggesting this yesterday, but I thought a bit more about the corruptedTCB testing process.

We should remove the TCB corruption ability via the skr.sh. So, rather than passing it via the skr.sh (or via an env. variable), we can add a var in line 23-27 

	// for testing mis-matched TCB versions allowTestingWithMismatchedTCB 
	// and MismatchedTCB 
        var allowTestingWithMismatchedTCB = false
	var MismatchedTCB = "ffffffff"

then in your code, you can check if the allowTestingWithMismatchedTCB is true, then use the MismatchedTCB as a value for the Tcbm. 

Finally, we can introduce an skr_test.sh script (variation of the current skr.sh) that passes an allowTestingWithMismatchedTCB flag (as opposed to the corrupted value) to the binary that overrides the allowTestingWithMismatchedTCB. (In essence, rename the current CorruptedTcbm variable to allowTestingWithMismatchedTCB variable and modify the codebase to expect this as a boolean flag) 

This way, the skr.sh (production entry point) will not have means to test while the skr_test.sh will have means to test. Then, we will need to add skr_test.sh to the container image.

@KenGordon thoughts on this?

KenGordon:
> I apologize for not suggesting this yesterday, but I thought a bit more about the corruptedTCB testing process.
> 
> We should remove the TCB corruption ability via the skr.sh. So, rather than passing it via the skr.sh (or via an env. variable), we can add a var in line 23-27
> 
> ```
> // for testing mis-matched TCB versions allowTestingWithMismatchedTCB 
> // and MismatchedTCB 
>     var allowTestingWithMismatchedTCB = false
> var MismatchedTCB = "ffffffff"
> ```
> 
> then in your code, you can check if the allowTestingWithMismatchedTCB is true, then use the MismatchedTCB as a value for the Tcbm.
> 
> Finally, we can introduce an skr_test.sh script (variation of the current skr.sh) that passes an allowTestingWithMismatchedTCB flag (as opposed to the corrupted value) to the binary that overrides the allowTestingWithMismatchedTCB. (In essence, rename the current CorruptedTcbm variable to allowTestingWithMismatchedTCB variable and modify the codebase to expect this as a boolean flag)
> 
> This way, the skr.sh (production entry point) will not have means to test while the skr_test.sh will have means to test. Then, we will need to add skr_test.sh to the container image.
> 
> @KenGordon thoughts on this?

yes, having a different script that lets an arm template very obviously control that it is testing a thing seems better.

svolos (approved):

PR #26, opened 2023-04-11
incorporating aasp code into this repo

stevendongatmsft:
This PR is for incorporating aasp repo into the SKR repo


cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+		return nil, fmt.Errorf("failed to read public key file %v", keyInfo.PublicKeyPath)
+	}
+	block, _ := pem.Decode([]byte(pubpem))
+	key, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("invalid public key in %v, error: %v", path, err)
+	}
+
+	var ciphertext []byte
+	if pubkey, ok := key.(*rsa.PublicKey); ok {
```
BryceDFisher, line 180: Are any other key types supported, like ECDSA?

cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+
+	return annotationBytes, nil
+}
+
+func (s *server) WrapKey(c context.Context, grpcInput *keyprovider.KeyProviderKeyWrapProtocolInput) (*keyprovider.KeyProviderKeyWrapProtocolOutput, error) {
+	var input keyProviderInput
+	str := string(grpcInput.KeyProviderKeyWrapProtocolInput)
+	err := json.Unmarshal(grpcInput.KeyProviderKeyWrapProtocolInput, &input)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Ill-formed key provider input: %v. Error: %v", str, err.Error())
```
BryceDFisher, line 200: I'm not sure how it works with grpc, but the newer way of handling errors is to wrap them (via an explicit type or fmt or errors) and only include text at this layer that is relevant and allow the inner error to convey its own message.

cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+	log.Printf("Key provider input: %v", input)
+
+	var dc = input.KeyUnwrapParams.Dc
+	if len(dc.Parameters["attestation-agent"]) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "attestation-agent must be specified in decryption config parameters: %v", str)
+	}
+	aa, _ := base64.StdEncoding.DecodeString(dc.Parameters["attestation-agent"][0])
+	log.Printf("Attestation agent name: %v", string(aa))
+
+	if string(aa) != "aasp" && string(aa) != "aaa" {
```
BryceDFisher: Now is a good chance to fully remove the "aaa" name since it was deprecated in favor of AASP.

BryceDFisher: In the request, is it case sensitive?  Would it make sense to do something like `!strings.EqualFold(aa,"aasp")`

cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+	var annotation AnnotationPacket
+	err = json.Unmarshal(annotationBytes, &annotation)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Ill-formed annotation packet: %v. Error: %v", input.KeyUnwrapParams.Annotation, err.Error())
+	}
+	log.Printf("Annotation packet: %v", annotation)
+
+	// bearerToken := ""
+
+	// // Note: obtaining access token through federated token file is AKS specific
```
BryceDFisher: And only if workload identities are implemented.  If the customer configures cluster identity or pod identity (deprecated) the below doesn't apply.

cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+		APIVersion: "api-version=2020-10-01",
+	}
+
+	skrKeyBlob := skr.KeyBlob{
+		KID:       annotation.Kid,
+		Authority: maa,
+		AKV:       mhsm,
+	}
+
+	if retrieveVCertChain(azure_info.CertCache, &EncodedUvmInformation); err != nil {
```
BryceDFisher, line 285: I'm not seeing where azure_info ever gets set to anything.

cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	return &keyprovider.KeyProviderKeyWrapProtocolOutput{
+		KeyProviderKeyWrapProtocolOutput: protocolBytes,
+	}, nil
+}
+
+// This logic should really belong to attest package. But because we will eventually change the way we retrieve certChain,
+// we should avoid putting temporary logic in attest pkg. Instead, just add final logic in attest and remove from here.
+func retrieveVCertChain(certCache attest.CertCache, encodedUvmInformation *common.UvmInformation) error {
```
BryceDFisher: This can be removed in favor of Heather's PR updating the cert cache

cmd/aasp/main.go
```diff
@@ -0,0 +1,468 @@
...
+	vcekCertChain, err := certCache.GetCertChain(SNPReport.ChipID, SNPReport.ReportedTCB)
+	if err != nil {
+		return errors.Wrapf(err, "retrieving cert chain from CertCache endpoint failed")
+	}
+
+	encodedUvmInformation.CertChain = string(vcekCertChain)
+	return nil
+}
+
+func (s *server) GetReport(c context.Context, in *keyprovider.KeyProviderGetReportInput) (*keyprovider.KeyProviderGetReportOutput, error) {
```
BryceDFisher, line 336: This can also just use the attest package

pkg/common/crypto.go
```diff
@@ -75,3 +79,60 @@ func PrivateKeyFromPEM(privatePEMString string) (*rsa.PrivateKey, error) {
 		return privateWrappingKey, nil
 	}
 }
+
+func RSAPrivateKeyFromJWK(jwKey *jwk.Key) (*rsa.PrivateKey, error) {
```
BryceDFisher, line 84: Does it make sense to support other key types than RSA?  You could return a https://pkg.go.dev/crypto#PrivateKey to more generically handle key types in other functions.  JWK even includes a key type parameter.  The JWT library you're importing has a JWKS extension that could help you do this as well: https://github.com/MicahParks/keyfunc.

cmd/aasp/main.go
stevendongatmsft: Do you have a concrete example of the error handling you mentioned above? 
Anyway, I personally think we should stick with the way grpc error handling here because it's very clear and more specific. Below is a sample error message produced by the status package. It's very clear that the error came from rpc and the error code indicates the type of rpc error and exactly what input caused the error as well. We do have control over description. 
`rpc error: code = InvalidArgument desc = Ill-formed key provider input: aaaa. Error: self defined error`

cmd/aasp/main.go
stevendongatmsft: yeah, we can do this.

cmd/aasp/main.go
stevendongatmsft: I don't think we currently supports ECDSA. I just looked up the types of keys AKV and HSM supports and it does look like EC  keys are supported. Let me reach out to Ken and Stavros and see why currently there is no impl on support EC keys.

stevendongatmsft:
Talked with Stavros regarding extending support for additional types of keys. Since there has not been a case requiring them, it should not be difficult to extend the current akv/skr package to support these keys. But Imo this should be a separate issue from a different PR.

examples/aasp/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+
+Even though `GUEST_IMAGE_MEASUREMENT` and `WORKLOAD_MEASUREMENT` are optional,
+we strongly recommend them to be included in the key release policy for
+better protection of the MHSM key which in turn protects your secrets.
+
+### Create an asymmetric key in MHSM and produce files related to the key
+
+This is a one time effort. Once the key is created, it can be used to protect
+as many secrets as possible.
```
BryceDFisher: `as desired`?

examples/aasp/README.md
```diff
@@ -0,0 +1,165 @@
...
+This guide provides a step-by-step instructions for the steps above. By following these instructions, users can ensure that their secrets are securely stored and can be used by the application. 
+This guide assumes a working mHSM and maa endpoint. See [prerequisites](#Prerequisites). 
+
+# Prerequisites:
+Before performing secret provisioning using AASP container, ensure that the following prerequisites are met:
+
+Users must have a functioning mHSM to store the private key and its associated release policy. 
+The release policy describes the conditions the key release request must meet in order for the mHSM to release the key. 
+
+To set up a MHSM instance, follow instructions [here with Azure CLI](https://learn.microsoft.com/en-us/azure/key-vault/managed-hsm/quick-create-cli),
```
BryceDFisher: Please pick either mHSM or MHSM.

examples/aasp/README.md
```diff
@@ -0,0 +1,165 @@
...
+You can follow the installation instructions for Azure CLI [here](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli).
+
+Users must have a working maa endpoint to interact with the mHSM.
+
+Users must build their own aasp container image and an image that allows users to issue grpcurl commands. 
+These two images can be built using the existing Dockerfiles: [Dockerfile.build](../../docker/aasp/Dockerfile.build) and [Dockerfile.sample](../../docker/aasp/Dockerfile.sample) respectively. 
+To build the images, run the following commands:
+
+```
+docker build -f Dockerfile.build .
```
BryceDFisher: Which directory do we run this from?

examples/aasp/README.md
stevendongatmsft: done

examples/aasp/README.md
stevendongatmsft: done

examples/aasp/README.md
```diff
@@ -0,0 +1,165 @@
...
+# Secret Provisioning Workflow:
+
+## Generate a key pair and prepare for the key release policy:
+
+This is a one time effort.
+Once the key is created, it can be used to protect as many secrets as possible. 
+The `setup-key-mhsm.sh` script creates a private/public key pair in the keyvault and a key release policy with the file name `keyname-release-policy.json`. 
+The public key is then downloaded as `keyname-pub.pem` to be used for wrapping the secret. 
+The script also assigns necessary access role(Managed HSM Crypto User) to the Managed Service Idenitty and it also generates a key info file that is later used for unwrapping the secret. 
+Depending on the operating system, make sure the end of line sequence(LF vs CRLF) is set correctly before running the script.
```
BryceDFisher, line 62: Are one of these required, or is it system dependent?

examples/aasp/README.md
stevendongatmsft: Done

examples/aasp/README.md
```diff
@@ -0,0 +1,165 @@
...
+## Generate a key pair and prepare for the key release policy:
+
+This is a one time effort.
+Once the key is created, it can be used to protect as many secrets as possible. 
+The `setup-key-mhsm.sh` script creates a private/public key pair in the keyvault and a key release policy with the file name `keyname-release-policy.json`. 
+The public key is then downloaded as `keyname-pub.pem` to be used for wrapping the secret. 
+The script also assigns necessary access role(Managed HSM Crypto User) to the Managed Service Idenitty and it also generates a key info file that is later used for unwrapping the secret. 
+Depending on the operating system, make sure the end of line sequence(LF vs CRLF) is set correctly before running the script. 
+
+In `setup-key-mhsm.sh` script, even though `GUEST_IMAGE_MEASUREMENT` and `WORKLOAD_MEASUREMENT` are optional, we strongly recommend them to be included in the key release policy for better protection of the MHSM key which in turn protects your secrets.
```
BryceDFisher: If you're going to recommend `GUEST_IMAGE_MEASUREMENT` how do you get it?

examples/aasp/README.md
```diff
@@ -0,0 +1,165 @@
...
+```bash
+# Create a managed identity for accessing MHSM. Note the Principle ID of the identity
+az identity create -g <resource-group-name> -n <identity-name>
+export MANAGED_IDENTITY=<principle-id>
+# Choose a MAA instance for the attestation service, e.g. e.g. sharedeus2.eus2.attest.azure.net
+export MAA_ENDPOINT=<maa-endpoint>
+# Specify a trusted measurement for Kata guest image
+export GUEST_IMAGE_MEASUREMENT=<sevsnp-vm-launch-measurement>
+# Specify a trusted measurement for the containers running in a confidential pod
+export WORKLOAD_MEASUREMENT=<hash-of-security-policy-for-confidential-pod>
```
BryceDFisher: Is security policy available at this point?

examples/aasp/README.md
```diff
@@ -0,0 +1,165 @@
...
+# Running the AASP container in a TEE:
+Use the AASP container to unwrap the secret with the private key stored in mHSM. 
+This step ensures that the private key is securely retrieved in AASP container and used for unwrapping the wrapped secret. 
+Before unwrapping the secret, the images for AASP container and another container that has grpcurl command installed for querying the aasp container need to be built, see [prerequisites](#Prerequisites). 
+
+Check [here](https://github.com/container-investigations/kata-verity/tree/kata-cc-based/katacc-bootstrap)
+for instructions on setting up an environment for running AASP container. 
+The instruction sets up an Azure VM, installs and configures a single node Kubernetes cluster, and sets up the network and kata runtimeClass in it. 
+Once you can run a sample pod with SEV-SNP support and cloud-api-adaptor. 
+You can run a sample deployment of a confidential pod with the AASP container and a container that can invokes the secret provisioning API of the AASP container. Use the following [sample pod yaml file](aasp-sample.yaml) to run the pod.
```
BryceDFisher: Does this work in AKS today?

examples/aasp/testkey000-pub.pem
BryceDFisher, line 1: Should probably delete this from the repo if the private key isn't available.

examples/aasp/testkey000-info.json
BryceDFisher, line 1: This looks like its configuring the dev mhsm.

docker/aasp/Dockerfile.build
```diff
@@ -0,0 +1,29 @@
+ARG GOLANG_VERSION=1.17
```
BryceDFisher: You updated to go 1.19 in the go.mod file.

examples/aasp/testkey000-pub.pem
stevendongatmsft: yeah, I deleted it, have not pushed yet.

docker/aasp/Dockerfile.build
stevendongatmsft: done

examples/aasp/testkey000-info.json
stevendongatmsft: This is used for wrapping the secret. It contains the maa endpoint info and these info are passed to the wrapped secret annotation.

examples/aasp/README.md
stevendongatmsft: I'm not sure what you mean. These are the things users need to make sure before they move onto the secret provisioning.

examples/aasp/README.md
stevendongatmsft: This instruction is for running aasp container on a VM with a k8s node. I believe this is different from AKS cluster.

docker/aasp/Dockerfile.build
stevendongatmsft: done

examples/aasp/README.md
stevendongatmsft: For now, not really so I think I'm going to comment out guest image and workload measurement for now. We can add later.

examples/aasp/README.md
stevendongatmsft: For now, not really so I think I'm going to comment out guest image and workload measurement for now. We can add later.

cmd/aasp/main.go
BryceDFisher: This comment is not correct.  Any environment can login via federated tokens if you provide the values below.

examples/aasp/README.md
```diff
@@ -0,0 +1,171 @@
...
+Once the key is created, it can be used to protect as many secrets as desired. 
+The `setup-key-mhsm.sh` script creates a private/public key pair in the keyvault and a key release policy with the file name `keyname-release-policy.json`. 
+The public key is then downloaded as `keyname-pub.pem` and a key info file is generated. 
+Both the public key and the key info file are used for wrapping the secret. 
+The script also assigns necessary access role(Managed HSM Crypto User) to the Managed Service Idenitty.
+Depending on the operating system, make sure the end of line sequence(LF vs CRLF) is set correctly before running the script. 
+
+`MANAGED_IDENTITY` has the following format: 
+```
+/subscriptions/subscription-id/resourceGroups/resource_group_name/providers/Microsoft.ManagedIdentity/userAssignedIdentities/msi
```
takuro-sato: I think it's easier to read if we use `<>` for the string which should be replaced like:
```
/subscriptions/<subscription-id>/resourceGroups/<resource_group_name>/providers/Microsoft.ManagedIdentity/<userAssignedIdentities>/msi
```

examples/aasp/README.md
```diff
@@ -0,0 +1,171 @@
...
+You can run a sample deployment of a confidential pod with the AASP container and a container that can invokes the secret provisioning API of the AASP container. Use the following [sample pod yaml file](aasp-sample.yaml) to run the pod. But create an image pull secret first with the following command. 
+
+```
+kubectl create secret docker-registry <secret-name> \
+    --namespace <namespace> \
+    --docker-server=<REGISTRY_NAME>.azurecr.io \
+    --docker-username=<appId> \
+    --docker-password=<password>
+```
+Replace the <image> and <imagePullSecrets> with your own build image and image pull secret.
```
takuro-sato: Is it possible to let the user pass these parameter without editing the file?

examples/aasp/README.md
```diff
@@ -0,0 +1,171 @@
...
+Response headers received:
+content-type: application/grpc
+
+Response contents:
+{
+  "KeyProviderKeyWrapProtocolOutput": "eyJrZXl3cmFwcmVzdWx0cyI6eyJhbm5vdGF0aW9uIjpudWxsfSwia2V5dW53cmFwcmVzdWx0cyI6eyJvcHRzZGF0YSI6IlQyTmxZVzV6SUdGeVpTQm1kV3hzSUc5bUlIZGhkR1Z5RFFwSWIzSnpaWE1nYUdGMlpTQTBJR3hsWjNNTkNnPT0ifX0="
+}
+
+```
+Base64 decode the result to see the encrypted secret.
```
takuro-sato, line 182: Can we add a script to deploy and test AASP for future PRs?
```bash
$ ./deploy_and_test.sh <resource_group> <and any other parameters> ....
OK
```
deploy_and_test.sh would do everything that this README.md tells (like creating mhsm instance, building docker image, and check the result of unwrap key command).

I am aiming to add an equivalent for attestation-container.

https://github.com/microsoft/confidential-sidecar-containers/blob/2fe9a2d62a7f5e61e57f5101562503580bb42966/docker/attestation-container/deploy_and_test.sh

I would be time consuming to write but it should help us for future PRs (Also it should help to add CI).

examples/aasp/README.md
```diff
@@ -0,0 +1,171 @@
...
+The release policy describes the conditions the key release request must meet in order for the mhsm to release the key. 
+
+To set up a mhsm instance, follow instructions [here with Azure CLI](https://learn.microsoft.com/en-us/azure/key-vault/managed-hsm/quick-create-cli),
+or through the [Azure portal](https://ms.portal.azure.com/#view/HubsExtension/BrowseResource/resourceType/Microsoft.KeyVault%2FmanagedHSMs). 
+You can follow the installation instructions for Azure CLI [here](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli).
+
+Users must have a working maa endpoint to interact with the mhsm.
+
+Users must build their own AASP container image and an image that allows users to issue grpcurl commands. 
+These two images can be built using the existing Dockerfiles: [Dockerfile.build](../../docker/aasp/Dockerfile.build) and [Dockerfile.sample](../../docker/aasp/Dockerfile.sample) respectively.
```
takuro-sato: Could you add explanation about what `Dockerfile.sample` is for? Also it might be worth to mention that its container is called "sample-unwrapped container" in this README.md.

pkg/aasp/keyprovider/keyprovider.proto
```diff
@@ -0,0 +1,37 @@
...
+
+message keyProviderGetReportOutput {
+  string reportHexString = 1;
+}
+
+service KeyProviderService {
+    rpc WrapKey(keyProviderKeyWrapProtocolInput) returns (keyProviderKeyWrapProtocolOutput) {};
+    rpc UnWrapKey(keyProviderKeyWrapProtocolInput) returns (keyProviderKeyWrapProtocolOutput) {};
+    rpc GetReport(keyProviderGetReportInput) returns (keyProviderGetReportOutput) {};
+    rpc SayHello (HelloRequest) returns (HelloReply) {}
```
takuro-sato, line 27: This is out of scope of this PR, but now we are getting similar endpoints with slightly different interface and name (attestation-container is adding a feature to fetch report as well for example).

Maybe we need to make them consistent in future (with same name, parameter, order of parameter, same status code, same output format, etc) among applications/protocols to minimize confusion.

examples/aasp/README.md
stevendongatmsft: good catch

examples/aasp/README.md
stevendongatmsft: I don't want to put them on the same line because that would be too long. Do you have any recommendations? 
This is copied and pasted from an official Azure documentation -> (https://learn.microsoft.com/en-us/azure/aks/hybrid/deploy-azure-container-registry)

examples/aasp/README.md
stevendongatmsft: yeah, this is updated

pkg/aasp/keyprovider/keyprovider.proto
stevendongatmsft: ok, we can definitely discuss in the future.

pkg/attest/certcache.go
```diff
@@ -166,7 +166,7 @@ func (certFetcher CertFetcher) retrieveCertChain(chipID string, reportedTCB uint
 		case "LocalTHIM":
 			uri = fmt.Sprintf(LocalTHIMUriTemplate, certFetcher.Endpoint)
 			// local THIM cert cache endpoint returns THIM Certs object
-			httpResponse, err := common.HTTPGetRequest(uri, false)
+			httpResponse, err := common.HTTPGetRequest(uri, true)
```
takuro-sato: Just question, but could you give me the context of this change?
(Why it's necessary, etc)

cmd/aasp/main.go
```diff
@@ -0,0 +1,494 @@
...
+	log.Printf("Decoded annotation: %v", string(annotationBytes))
+
+	var annotation AnnotationPacket
+	err = json.Unmarshal(annotationBytes, &annotation)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Ill-formed annotation packet: %v. Error: %v", input.KeyUnwrapParams.Annotation, err.Error())
+	}
+	log.Printf("Annotation packet: %v", annotation)
+
+	// bearerToken := ""
```
takuro-sato, line 275: nit: How about removing these comment to make it look tidy?

If comment out makes sense for this case, it might be better to add comment to explain it.

pkg/msi/msi.go
```diff
@@ -0,0 +1,32 @@
...
+	"os"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
+)
+
+// Convert federated token stored in a file into access token for a resource
+// Borrowed from https://github.com/Azure/azure-workload-identity/blob/c155ecee0d9fa681c15ead4bbdce729fd8c99da1/pkg/proxy/proxy.go#L195
+// See tutorial at: https://learn.microsoft.com/en-us/azure/aks/learn/tutorial-kubernetes-workload-identity
+
+func GetAccessTokenFromFederatedToken(ctx context.Context, federatedTokenFile, clientID, tenantID, resource string) (string, error) {
```
takuro-sato, line 15: Can we add test for it?

pkg/common/crypto.go
takuro-sato, line 84: Can we add test for this function?

pkg/common/crypto.go
```diff
@@ -75,3 +79,60 @@ func PrivateKeyFromPEM(privatePEMString string) (*rsa.PrivateKey, error) {
...
 	}
 }
+
+func RSAPrivateKeyFromJWK(jwKey *jwk.Key) (*rsa.PrivateKey, error) {
+	jwkJSONBytes, err := json.Marshal(jwKey)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Released key cannot be marshalled into bytes: %v", err)
+	}
+
+	//TODO: remove DP, DQ and QI because they are not needed for decryption
```
takuro-sato: nit: It looks trivial so can we do it in this PR?

cmd/aasp/main.go
stevendongatmsft: Hi Yeah, I will add comment to explain why commenting this out for now is necessary.

examples/aasp/README.md
stevendongatmsft: This is done

pkg/attest/certcache.go
stevendongatmsft: So this is for getting the certificate from the THIM endpoint. As far as I know, the THIM endpoint is local so we use http instead of https. But I think I'm going to reverse this change because there is no environment that allows us to use the THIM endpoint right now.

pkg/common/crypto.go
stevendongatmsft: yeah, this is removed.

pkg/attest/certcache.go
KenGordon: I suspect this ought to go in a separate PR, along with the https->http in the template.

examples/aasp/README.md
stevendongatmsft: In order to automate everything, the following has to be done by the scripts:

1. The script will make MAA and mHSM endpoint required and reports error if not.

2. create and import keys, which can be easily done by the setup-key-mhsm.sh script.
3. wrapping a secret using the aasp binary as a command line tool
4. build and push docker images with wrapped secret and a script that does the unwrapping copied into the container
5. setup a nested k8s cluster with /dev/sev support
6. create k8s secret for container registry access.
7. deploy the example and make the unwrap scripts automatically run to unwrap the secrets.

Step 5 is currently under development and the way we are testing aasp container is through using peer pod approach. So this won't be 100% complete until we have a full nested Kata K8s environment available. (Currently a dependency outside of our team) I suggest that I can do this on a separate PR and use an issue to track it. But yeah, eventually this script would be beneficial to customers.

pkg/attest/certcache.go
stevendongatmsft: Yeah this is removed for now

pkg/aasp/keyprovider/keyprovider.proto
BryceDFisher: We should really deal with this while both PRs are open so that we don't end up with backwards compatibility issues. @takuro-sato

docker/aasp/Dockerfile.build
```diff
@@ -0,0 +1,29 @@
...
+
+RUN apk update && apk --no-cache add curl
+
+RUN mkdir -p bin
+
+COPY --from=build /aasp ./bin/
+COPY --from=build /get-snp-report ./bin/
+COPY --from=build /get-fake-snp-report ./bin/
+COPY --from=build /verbose-report ./bin/
+RUN date > /made-date
```
BryceDFisher: I not a huge fan of putting the build date-time into the container because it prevents reproducible builds by customers.

pkg/msi/msi.go
stevendongatmsft: This relies on k8s cluster mount a volume for federated token file. I don't think it's feasible to add a test for this piece of code.

pkg/common/crypto.go
stevendongatmsft: I've added test for this.

docker/aasp/Dockerfile.build
stevendongatmsft: This is removed.

pkg/aasp/keyprovider/keyprovider.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message HelloRequest {
+  string name = 1;
```
hgarvison, line 31: What's the point of making separate messages just for testing? I would remove the "Hello" req/rep stuff and just use the key/report messages for the grpc test since test code shouldn't be in production code.

pkg/attest/maa.go
```diff
@@ -88,15 +92,39 @@ func newAttestSNPRequestBody(snpAttestationReport []byte, vcekCertChain []byte,
...
+	// }
+	var maaReportJSONBytes []byte
+
+	if len(uvmReferenceInfo) != 0 {
+		maaReportWithEndorsement := maaReportWithEndorsement{
+			maaReport: &maaReport{
+				SNPReport: base64.URLEncoding.EncodeToString(snpAttestationReport),
+				CertChain: base64.URLEncoding.EncodeToString(vcekCertChain),
+			},
+			Endorsements: base64urlEncodedmaaEndorsement,
```
hgarvison, line 113: Instead of this if-else, I would just set base64urlEncodedmaaEndorsement to the base64 encoding of the byte string (line82), since I believe this will be "" if the uvmInfo == "".  Then you can create the maaReport object, setting Endorsements equal to the variable regardless of if you update the base64url endorsement var. Otherwise you have a lot of duplicated code in this if-else.

pkg/common/crypto.go
```diff
@@ -75,3 +80,56 @@ func PrivateKeyFromPEM(privatePEMString string) (*rsa.PrivateKey, error) {
...
+		P string `json:"p"`
+		Q string `json:"q"`
+	}
+
+	if err := json.Unmarshal(jwkJSONBytes, &jwkData); err != nil {
+		return nil, errors.Wrapf(err, "Released key is not a RSA private key.")
+	}
+	n, err := base64.RawURLEncoding.DecodeString(jwkData.N)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Interpretting jwk key element failed")
```
hgarvison, line 103: I would make the error more specific about which key element failed (N in this case) for all these url encodings.

examples/aasp/README.md
hgarvison: I would just specify the use of LF is set before running the script, since it can only be run in WSL in Windows or just make a powershell script that does the same stuff for the Windows users.

docker/aasp/Dockerfile.sample
```diff
@@ -0,0 +1,17 @@
+FROM alpine:3.17.1
```
hgarvison, line 1: I would like a short README here on how these two dockerfiles relate/can be used,

PR #27, opened 2023-04-21
Ken/sev guest

KenGordon:
In case Heather's equivalent fix does not make it in.


KenGordon:
Seems to be to many changes

PR #28, opened 2023-04-24
Check for sev-guest dev (6.1)

KenGordon:
The PSP device is exposed as /dev/sev-guest rather than /dev/sev. When determining whether this is a real SNP box we need to check both. If not a real SNP box then a fake report is generated to allow for testing and development.

### DIFF
--- a/pkg/attest/attest.go
+++ b/pkg/attest/attest.go
@@ -18,11 +18,15 @@ import (
 // RawAttest returns the raw attestation report in hex string format
 func RawAttest(inittimeDataBytes []byte, runtimeDataBytes []byte) (string, error) {
 	// check if sev device exists on the platform; if not fetch fake snp report
-	var fetchRealSNPReport bool
-	if _, err := os.Stat("/dev/sev"); os.IsNotExist(err) {
-		fetchRealSNPReport = false
-	} else {
-		fetchRealSNPReport = true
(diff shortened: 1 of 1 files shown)


pkg/attest/attest.go
```diff
@@ -51,11 +51,14 @@ func Attest(maa MAA, runtimeDataBytes []byte, uvmInformation common.UvmInformati
...
 
 	// check if sev device exists on the platform; if not fetch fake snp report
-	var fetchRealSNPReport bool
-	if _, err := os.Stat("/dev/sev"); os.IsNotExist(err) {
-		fetchRealSNPReport = false
-	} else {
-		fetchRealSNPReport = true
+
+	fetchRealSNPReport := true
+	if _, err := os.Stat("/dev/sev"); errors.Is(err, os.ErrNotExist) {
```
hgarvison, line 60: This check should be added to the RawAttest function as well

svolos: Good catch. that's line 22

svolos (changes requested):
RawAttest requires the same check as Attest

pkg/attest/attest.go
KenGordon: Done

hgarvison (approved):

svolos (approved):

svolos:
These changes have been already merged as part of PR #24

PR #30, opened 2023-05-03
Add attestation container

takuro-sato:
## Changes
- Copy attestation container to this repo from CCF ([961c11e](https://github.com/microsoft/confidential-sidecar-containers/pull/30/commits/961c11e77b1b248c3c15038f3c4ae6a11f983f31))

```
This commit is to copy contents of https://github.com/takuro-sato/CCF/tree/335b7101a5616c67fa796ad7fd74e9711b67314b/attestation-container
to this repository with the commit history.

Pre-requisite:
- Install git-filter-repo (https://github.com/newren/git-filter-repo).
  Using this tool seems the easiest way when you want to move multiple directories
  (cosesign1 and sign1util) to a new repo.
Inside hcsshim/:
- git clone <fork of CCF> (It should be newly cloned)
- git filter-repo --path attestation-container
- git checkout -b filter-attestation-container
- git push origin filter-attestation-container
Inside this repository:
- git remote add ccf <fork of CCF>
- git merge --allow-unrelated-histories ccf/filter-attestation-container
```
- Change dir structure so that it marches with existing dir structure
- Update go.mod because of attestation-container's dependencies
  - Run `go mod tidy`
  - Update required go version. It's under discussion [here](https://github.com/microsoft/confidential-sidecar-containers/pull/30#discussion_r1184676261)
 - Run `go mod vendor`
- [Rename 'attestation report endorsement' to 'platform certificate' for consistency](https://github.com/microsoft/confidential-sidecar-containers/pull/30/commits/4dc3ee2e2856421893b4ed7b014c41db9b651ede)
- - [x] Refactor `pkg/attest` and other packages. Remove duplication

### Manual test inside Confidential ACI instance:
```bash
# Prepare for E2E test
go run go run ./cmd/attestation-container/ 

# In another terminal
go test ./...
```

### Deploy to Confidential ACI then run test

This should be in the CI in future

With normal Azure account:
```bash
az login
az account set --subscription <your subscription>
./docker/attestation-container/deploy_and_test.sh  <AZURE_RESOURCE_GROUP> <DOCKER_REGISTRY_NAME> <DEPLOY_ID (anything)>
```

With service principle:
```bash
./docker/attestation-container/login_service_principal.sh <AZURE_TENANT_ID> <AZURE_APP_ID> <AZURE_SERVICE_PRINCIPAL_PASSWORD> <AZURE_SUBSCRIPTION_ID>
./docker/attestation-container/deploy_and_test.sh  <AZURE_RESOURCE_GROUP> <DOCKER_REGISTRY_NAME> <DEPLOY_ID (anything)>
```

### Test existing code affected by the refactoring
```
go run ./cmd/skr

# in another terminal
export AttestClientRuntimeData="eyJrZXlzIjpbeyJlIjoiQVFBQiIsImtleV9vcHMiOlsiZW5jcnlwdCJdLCJraWQiOiJOdmhmdXEyY0NJT0FCOFhSNFhpOVByME5QXzlDZU16V1FHdFdfSEFMel93Iiwia3R5IjoiUlNBIiwibiI6InY5NjVTUm15cDh6Ykc1ZU5GdURDbW1pU2VhSHB1akcyYkNfa2VMU3V6dkRNTE8xV3lyVUp2ZWFhNWJ6TW9PMHBBNDZwWGttYnFIaXNvelZ6cGlORExDbzZkM3o0VHJHTWVGUGYyQVBJTXUtUlNyek41NnF2SFZ5SXI1Y2FXZkhXay1GTVJEd0FlZnlOWVJIa2RZWWtnbUZLNDRoaFVkdGxDQUtFdjVVUXBGWmp2aDRpSTlqVkJkR1lNeUJhS1FMaGpJNVdJaC1RRzZaYTVzU3VPQ0ZNbm11eXV2TjVEZmxwTEZ6NTk1U3MtRW9CSVktTmlsNmxDdHZjR2dSLUlialVZSEFPczVhamFtVHpnZU84a3gzVkNFOUhjeUtteVVac2l5aUY2SURScDJCcHkzTkhUakl6N3Rta3BUSHg3dEhuUnRsZkUyRlV2MEI2aV9RWWxfWkE1USJ9XX0="
export AttestClientMAAEndpoint=<Your own MAA endpoint>
bash attest_client.sh
bash ./docker/skr/tests/attest_client.sh
```

## TODOs for future PRs
- Run attestation-container's test in CI. 
  - Create service principal with enough privilege, and put it as GitHub sercret
  - Control access to the new CI. e.g. only pre-registered member can run the CI https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization
 -  Support Linux kernel 6.1. discussion: https://github.com/microsoft/confidential-sidecar-containers/pull/30#discussion_r1183650544
 - As I discussed with Ken, we can add retry mechanism to accessing to MAA like fetching platform certs
 - Now there is to naming convention for test `Test_*` and `Test*`, I guess we can just `Test*` since it's golang's recommendation. I leave it for now since it makes the diff even more complicated


vendor/bazil.org/fuse/go.mod
```diff
@@ -1,12 +0,0 @@
-module bazil.org/fuse
-
-go 1.13
-
-require (
-	github.com/dvyukov/go-fuzz v0.0.0-20200318091601-be3528f3a813
```
KenGordon, line 6: Is this really gone?

cmd/attestation-container/attestation-container.go
```diff
@@ -0,0 +1,137 @@
...
+		log.Println("Serving virtual attestation report")
+		return &pb.FetchAttestationReply{}, nil
+	}
+
+	reportBytes, err := attest.FetchAttestationReportByte(reportData)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to fetch attestation report: %s", err)
+	}
+
+	var attestationEndorsement []byte
```
KenGordon: Can you try to make the names of these (the platform certificates) match with the names in the previous SKR code?
MAA uses endorsements to mean the uvm endorsement (which I call uvm reference info) so "endorsement" is likely to lead to confusion.

pkg/attest/attestation.go
```diff
@@ -0,0 +1,152 @@
...
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+// Data structures are based on SEV-SNP Firmware ABI Specification
+// https://www.amd.com/en/support/tech-docs/sev-secure-nested-paging-firmware-abi-specification
```
KenGordon: This whole file will need to change for 6.1 linux where the PSP interface device has changed.

KenGordon: If CCF is using this now we'll need to sort something out. We also need to decide whether to only release this for 6.1

vendor/bazil.org/fuse/go.mod
takuro-sato: Haven't checked in detail, but what I did so far is:

```bash
go mod edit -go=1.20 # since the attestation-container's code requires newer version
go mod tidy
go mod vendor
```

I'll double check before I open this PR

cmd/attestation-container/attestation-container.go
takuro-sato: Now we are trying to put it to this repo so it makes sense to have consistent naming across the repo rather than to make it consistent with CCF.
I'll do so.

vendor/bazil.org/fuse/go.mod
KenGordon: Doe we really need 1.20 here?
Maybe we should update the whole repo to require 1.20 or back this off to whatever version the rest of it needs. For sure we don't want to have a mix.

takuro-sato: > Doe we really need 1.20 here?

No, but I believe it's reasonable choice.
At least we need go1.17 or later since  attestation-container uses its feature, so upgrading go is necessary.
And if we need to upgrade go, I prefer to choose 1.20 because it's the latest (potentially we can reduce required go upgrades) and also attestation-container uses 1.20 currently (it means less changes on entire code base).
Also I am aiming to use ['-skip' option for `go test` command introduced in go 1.20](https://tip.golang.org/doc/go1.20) which should be useful for us to skip SNP dependent tests for some part of CI.

Potential down side of using latest is that users with older version of go might not be able to use our packages. But I don't think it's an actual problem because go is easy to upgrade version.

https://tip.golang.org/doc/godebug

> Go’s emphasis on backwards compatibility is one of its key strengths.

So I think if people need to use our packages, they will just upgrade their go runtime. 

> For sure we don't want to have a mix.

I agree that. I think it's inevitable when we want to keep everything in one module.

cmd/attestation-container/attestation-container.go
takuro-sato: It's done here [4dc3ee2](https://github.com/microsoft/confidential-sidecar-containers/pull/30/commits/4dc3ee2e2856421893b4ed7b014c41db9b651ede)

pkg/attest/attest.go
```diff
@@ -98,8 +82,16 @@ func RawAttest(inittimeDataBytes []byte, runtimeDataBytes []byte) (string, error
...
 func (certState *CertState) Attest(maa MAA, runtimeDataBytes []byte, uvmInformation common.UvmInformation) (string, error) {
+	inittimeDataBytes, err := base64.StdEncoding.DecodeString(uvmInformation.EncodedSecurityPolicy)
+	if err != nil {
+		return "", errors.Wrap(err, "decoding policy from Base64 format failed")
+	}
+	logrus.Debugf("   inittimeDataBytes:    %v", inittimeDataBytes)
+
 	// Fetch the attestation report
-	SNPReportBytes, inittimeDataBytes, err := GetSNPReport(uvmInformation.EncodedSecurityPolicy, runtimeDataBytes)
+	reportFetcher := NewAttestationReportFetcher()
```
takuro-sato: I guess there is no use case to use MAA with fake attestation report, so I am hard-coding to use `Real AttestationReportFetcher`, but let me know if fake one is needed here

pkg/attest/attestation_report_fetcher.go
```diff
@@ -0,0 +1,221 @@
...
+type AttestationReportFetcher interface {
+	// Fetches attestation report as []byte.
+	// reportData is guest-provided data defined in SEV-SNP Firmware ABI Specification.
+	FetchAttestationReportByte(reportData [REPORT_DATA_SIZE]byte) ([]byte, error)
+	// Fetches attestation report as hex.
+	// reportData is guest-provided data defined in SEV-SNP Firmware ABI Specification.
+	FetchAttestationReportHex(reportData [REPORT_DATA_SIZE]byte) (string, error)
+}
+
+func NewAttestationReportFetcher() AttestationReportFetcher {
```
takuro-sato, line 140: This interface ('New...' function that returns interface rather than struct) is based on design of golang's standard library
e.g. https://pkg.go.dev/crypto/sha256#New224
We can provide `UnsafeNewFakeAttestationReportFetcher` which returns fake report in the same way but with `hostData` as parameter.
With this method we don't have to include `hostData` in functions to fetch real attestation report which don't need it.
Users of this package can switch between `realAttestationReportFetcher` and `fakeAttestationReportFetcher` easily using dependency injection or similar techniques.

pkg/attest/attestation_report_fetcher.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	const SNP_REPORT_OFFSET = 32
+	reportBytes := reportRspBytes[SNP_REPORT_OFFSET : SNP_REPORT_OFFSET+ATTESTATION_REPORT_SIZE]
+	if common.GenerateTestData {
+		ioutil.WriteFile("snp_report.bin", reportBytes, 0644)
+	}
+	return reportBytes, nil
+}
+
+func (f *realAttestationReportFetcher) FetchAttestationReportHex(reportData [REPORT_DATA_SIZE]byte) (string, error) {
```
takuro-sato, line 183: This is the one that can be used instead of `RawAttest`

pkg/attest/platform_cert_fetcher_test.go
```diff
@@ -0,0 +1,140 @@
...
+			chipID:          ValidChipID,
+			platformVersion: ValidPlatformVersion,
+			expectedError:   errors.Errorf("invalid endpoint type: InvalidEndpointType"),
+			expectErr:       true,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			certchain, _, err := tc.certFetcher.GetCertChain(tc.chipID, tc.platformVersion)
```
takuro-sato, line 121: I changed it to test public API rather than private function (originally `retrieveCertChain` was tested instead) to avoid the case
where we don't realize that we break a public API.
Maybe we can just remove `retrieveCertChain` and put its contents inside `GetCertChain`.
Currently `GetCertChain` just calls `retrieveCertChain` and doesn't anything aside it.

pkg/attest/attest_test.go
takuro-sato, line 1: If I put every test here it will be too much, so I moved test cases to smaller files.

DomAyre (approved):
Looks okay to me once the conversations are resolved

pkg/attest/attest.go
DomAyre: I don't see that it would be needed, but you could check with the previous implementer?

cmd/attestation-container/attestation-container.go
```diff
@@ -0,0 +1,145 @@
...
+type server struct {
+	pb.AttestationContainerServer
+}
+
+func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRequest) (*pb.FetchAttestationReply, error) {
+	reportData := [attest.REPORT_DATA_SIZE]byte{}
+	if len(in.GetReportData()) > attest.REPORT_DATA_SIZE {
+		return nil, status.Errorf(codes.InvalidArgument, "`report_data` needs to be smaller than %d bytes. size: %d bytes", attest.REPORT_DATA_SIZE, len(in.GetReportData()))
+	}
+	copy(reportData[:], in.GetReportData())
```
BryceDFisher, line 44: Out of curiosity, is the copy used to avoid the nil pointer possibility from `GetReportData`?

cmd/attestation-container/attestation-container_test.go
```diff
@@ -0,0 +1,113 @@
...
+	// public key bytes in UTF-8 (https://go.dev/blog/strings)
+	publicKey := []byte("public-key-contents")
+	r, err := c.FetchAttestation(ctx, &pb.FetchAttestationRequest{ReportData: publicKey})
+	if err != nil {
+		log.Fatalf("could not get attestation. attestation-container needs to be run before run this test: %v", err)
+	}
+	// Verify attestation
+	attestation := r.GetAttestation()
+	if len(attestation) == 0 {
+		log.Fatalf("attestation is empty")
```
BryceDFisher: Wouldn't this more typically be `t.Fatal(...)`?

pkg/attest/attest.go
BryceDFisher: The only reason I would see to submit a fake SNP report to MAA is to test that they won't accept the report unless its properly signed.

pkg/attest/attestation_report_fetcher_snp_required_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//go:build !skip_snp_required
```
BryceDFisher, line 4: I like your use of build tags here to skip tests based on available hardware

pkg/common/info.go
```diff
@@ -31,44 +31,37 @@ type THIMCerts struct {
...
 }
 
-func (thimCerts *THIMCerts) GetLocalCerts(encodedHostCertsFromTHIM string) (string, uint64, error) {
-	var thimTcbm uint64
-	hostCertsFromTHIM, err := base64.StdEncoding.DecodeString(encodedHostCertsFromTHIM)
+func ParseTHIMCerts(base64EncodedHostCertsFromTHIM string) (THIMCerts, error) {
+	certificatesRaw, err := base64.StdEncoding.DecodeString(base64EncodedHostCertsFromTHIM)
 	if err != nil {
-		return "", thimTcbm, errors.Wrapf(err, "base64 decoding platform certs failed")
+		return THIMCerts{}, fmt.Errorf("Failed to decode ACI certificates: %s", err)
```
BryceDFisher: I think we should keep the error wrapping, but return the struct

vendor/bazil.org/fuse/go.mod
takuro-sato: As discussed, changed  the version so that it matches hcsshim (1.18 in go.mod and 1.19 for ci).

cmd/attestation-container/attestation-container.go
takuro-sato: The main purpose is to convert byte slice (`in.GetReportData()`) to byte array (`reportData`) so that it can be passed to `FetchAttestationReportByte()`.
I made `FetchAttestationReportByte()` take array rather than slice so that users of the function don't get length error in run time.

As you pointed out, it handles nil as well.

cmd/attestation-container/attestation-container_test.go
takuro-sato: That's true. Fixed in [ddc9ed7](https://github.com/microsoft/confidential-sidecar-containers/pull/30/commits/ddc9ed7fa0cd5258430761d4380be4c7f85cf5bc)

pkg/common/info.go
takuro-sato: Fixed in bce3342

pkg/attest/attest.go
KenGordon: To try this on a non SNP machine a report and matching certs from a real SNP box can be used.

takuro-sato: It seems useful for some use case, now it returns fake report if real one is no available after 5b3c1fe

docker/attestation-container/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# To build this:
+#   cd <project root directory> # It's necessary to copy contents of parent directory
+#   docker build -t attestation-container -f docker/attestation-container/Dockerfile . [--build-arg variant=dev]
+
+# Two variants: run (production) or dev (development)
+ARG variant=run
+
+# Build
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.19-fips-cbl-mariner2.0 AS builder-app
```
BryceDFisher, line 11: The go.mod got changed to go1.18, but the build is still using go1.19. Not sure if this was overlooked.

.github/workflows/ci.yml
```diff
@@ -81,10 +81,14 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.19
+        go-version: "1.19"
```
BryceDFisher, line 84: Also using go.19 here.

pkg/attest/platform_cert_fetcher.go
```diff
@@ -118,9 +125,79 @@ type CertFetcher struct {
...
+	return CertFetcher{
+		EndpointType: "AzCache",
+		Endpoint:     "global.acccache.azure.net",
+		TEEType:      "SevSnpVM",
+		APIVersion:   "api-version=2020-10-15-preview",
+	}
+}
+
+const (
+	defaultRetryBaseSec    = 2
```
stevendongatmsft: According to Ken's comment, I think the retry would be most useful when it comes to intermittent failures of the network. If this is the case, I would think a defaultRetryBaseSec of 2 is too small just because it's unlikely that the network hiccup would recover in a matter of 2 seconds. I would suggest to increase this to 6 or 10 seconds.

takuro-sato: @stevendongatmsft @KenGordon Updated

cmd/attestation-container/README.md
```diff
@@ -0,0 +1,65 @@
...
+```bash
+go build
+./attestation-container
+```
+
+## API
+
+The gPRC API is defined in [attestation-container.proto](https://github.com/microsoft/confidential-sidecar-containers/tree/main/cmd/attestation-container/protobuf/attestation-container.proto).
+
+Note that gPRC communication is used over [Unix domain sockets (UDS)](https://en.wikipedia.org/wiki/Unix_domain_socket). You can find an example client code in [the E2E test](https://github.com/microsoft/confidential-sidecar-containers/tree/main/cmd/attestation-container/attestation-container_test.go).
```
achamayou, line 45: It would be good to mention why Unix sockets are chosen here: because the security policy can explicitly enforce that an emptyDir mount is used, guaranteeing that only containers in the same SEV-SNP VM can access the attestation container.

It would be disastrously insecure to expose this gRPC server over network sockets, on which the security policy cannot assert anything: any random process with network access could obtain an attestation report with their user claims.

cmd/attestation-container/attestation-container.go
```diff
@@ -0,0 +1,145 @@
...
+	"google.golang.org/grpc/status"
+)
+
+var (
+	socketAddress             = flag.String("socket-address", "/tmp/attestation-container.sock", "The socket address of Unix domain socket (UDS)")
+	platformCertificateServer = flag.String("platform-certificate-server", "", "Server to fetch platform certificate. If set, certificates contained in security context directory are ignored. Value is either 'Azure' or 'AMD'")
+	insecureVirtual           = flag.Bool("insecure-virtual", false, "If set, dummy attestation is returned (INSECURE: do not use in production)")
+
+	platformCertificateValue *common.THIMCerts = nil
+	// UVM Endorsement (UVM reference info)
```
achamayou, line 31: A link to the UVM endorsement format documentation would be great here.

docker/attestation-container/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# pre-copy/cache go.mod for pre-downloading dependencies and only redownloading them in subsequent builds if they change
+COPY . .
+RUN go mod download && go mod verify
+
+# Note: Use CGO_ENABLED=0 to build statically-linked binaries deployable in scratch image
+RUN CGO_ENABLED=0 go build -v -o /usr/local/bin/app ./cmd/attestation-container/
+
+FROM builder-app AS builder-test
+WORKDIR /usr/src/app
+## Pre-compile tests to avoid I/O errors in ACI
```
achamayou, line 23: These errors have been fixed, the comment is stale.

achamayou (approved):

PR #34, opened 2023-05-10
Update to GA sidecar versions, make the encfs sidecar be privileged.

KenGordon:
Update to GA sidecar versions to *:2.2
Make the encfs sidecar be privileged so it can see all the required devices in /dev


stevendongatmsft (approved):
LGTM

stevendongatmsft:
I noticed that some test cases in Test_AKV began to fail. It all passed before but suddenly they began to fail. This makes me think that these certs have some external dependencies on Stavros' vault? Wondering whether it's possible that something was deleted or modified.

KenGordon:
> I noticed that some test cases in Test_AKV began to fail. It all passed before but suddenly they began to fail. This makes me think that these certs have some external dependencies on Stavros' vault? Wondering whether it's possible that something was deleted or modified.



> I noticed that some test cases in Test_AKV began to fail. It all passed before but suddenly they began to fail. This makes me think that these certs have some external dependencies on Stavros' vault? Wondering whether it's possible that something was deleted or modified.

For the moment I have commented these out. Takuro had mentioned this. I will ask him to replace the static certs a with self signed chain in a subsequent PR.

PR #35, opened 2023-05-10
Refactor and re-use code for setting up http server

stevendongatmsft:
There was a new requirement for Azure Attestation Secret Provisioning to implement http interfaces instead of purely relying on grpc protocol. This PR refactors code related to gin server that serves http requests so that AASP could also take advantage of existing `setupserver` code. 

I have already built and tested all http APIs after code refactoring and all work as expected.


pkg/http_listener/http_listener.go
```diff
@@ -0,0 +1,212 @@
...
+	jwkJSONBytes, err := json.Marshal(jwKey)
+	if err != nil {
+		c.JSON(http.StatusForbidden, gin.H{"error": err.Error()})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{"key": string(jwkJSONBytes)})
+}
+
+func SetupServer(certState *attest.CertState, identity *common.Identity, uvmInfo *common.UvmInformation) *gin.Engine {
```
takuro-sato: I guess sharing `SetupServer` between applications might potentially lead to some troubles in future because their requirements can be changed separately.
For example SKR might want to add an endpoint `/foo` which is not required by AASP. In that case we need to write code like `if application == "SKR" { server.GET("/foo", handleFoo)}`, and it's not tidy.
Also even a same endpoint can have slightly different requirement in future, it will be even more painful.

So how about trying to reuse only endpoint functions like `postKeyRelease`?
For now `SetupServer` would be duplicated but in a long run it would be a bit safer.

pkg/attest/attest.go
```diff
@@ -16,6 +16,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	ServerCertState *CertState
```
takuro-sato: I am wondering if we can avoid these global variables inside package.
I guess it's useful because gin's endpoint handlers only takes `gin.Context` like `func postKeyRelease(c *gin.Context)` and it's troublesome to pass additional parameters like `ServerCertState ` (Please let me know if it's not).

But I guess it's still possible with [`c.MustGet`](https://stackoverflow.com/questions/34046194/how-to-pass-arguments-to-router-handlers-in-golang-using-gin-web-framework) for example.
It's not ideal since it's not super straightforward but I believe it's still worth to do to avoid to have global variables which are specific to application rather than essential thing to the package itself.

pkg/http_listener/http_listener.go
```diff
@@ -0,0 +1,212 @@
+package http_listener
```
takuro-sato: The name sounds like a general package to handle http request.
Can we have a name that suggest what it actually offers?

takuro-sato (commented):
Thank you for creating PR! I put some comment so please let me know what you think

pkg/http_listener/http_listener.go
takuro-sato: I am thinking something like `gin-endpoints`

pkg/attest/attest.go
takuro-sato: Using `c.MustGet` might be not the only way, but if we use it, we can change code like this:

```
maaToken, err := attest.ServerCertState.Attest(maa, runtimeDataBytes, *common.EncodedUvmInformation)
```
->
```
serverState := c.MustGet("serverState").(*attest.CertState)
maaToken, err := serverState .Attest(maa, runtimeDataBytes, *common.EncodedUvmInformation)
```

Code to add (You can use it in main()):
```
func RegisterGlobalStates(certState *attest.CertState) gin.HandlerFunc {
    return func(c *gin.Context) {
        c.Set("databaseConn", certState)
        c.Next()
    }
}
```

main():
```
server.Use(RegisterGlobalStates(certState/*global variable defined in main.go*/))
```

pkg/http_listener/http_listener.go
takuro-sato, line 1: Sorry for adding comment later, but maybe we can make it `internal` so that we don't have to keep the API stable?
https://go.dev/doc/go1.4#internalpackages

Here is an example usage of `internal/`. https://github.com/microsoft/hcsshim/tree/main/internal

pkg/http_listener/http_listener.go
stevendongatmsft: Yes, different servers can have different requirements. So the refactoring here is to give different server types the ability to implement their own logics for extracting information used for key release(see UnderlyingServer interface). However, once these information gets extracted, the way we do secret release or getting maa token should be universal. 

Imo, all 4 methods postKeyRelease, postMAAAttest, postRawTest and getStatus all can be re-used because regardless the server type whether it's http or grpc, we will have all four types of APIs.

pkg/http_listener/http_listener.go
stevendongatmsft: It changed it to `gin`, what do you think?

pkg/attest/attest.go
stevendongatmsft: I changed the code based on your suggestion, please take a look.

pkg/http_listener/http_listener.go
stevendongatmsft: Hi Takuro, I'm not sure what this mean.

cmd/skr/main.go
```diff
@@ -334,5 +128,5 @@ func main() {
 		Tcbm:        thimTcbm,
 	}
 
-	setupServer(certState, info.Identity).Run(url)
+	server.SetupGinServer(&certState, &info.Identity, &EncodedUvmInformation, url)
```
takuro-sato, line 131: I saw a commit "[refactoring to make underlying server replaceable](https://github.com/microsoft/confidential-sidecar-containers/pull/35/commits/8b9396cc850cda84fbe0c28122dfc2017fac5340)".

If we want to make it easy to replace, should we rename it to `SetupServer` without 'Gin' in the name?

pkg/attest/attest.go
takuro-sato: Looks great. Can you do the same for `WorkloadIdentity` and `EncodedUvmInformation` in info.go as well please?

internal/httpginendpoints/httpginendpoints.go
```diff
@@ -0,0 +1,230 @@
...
+func GetStatus(c *gin.Context) {
+	if ready {
+		c.JSON(http.StatusOK, gin.H{"message": "Status OK"})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{"message": "Status NOT OK"})
+}
+
+// postRawAttest retrieves a hardware attestation report signed by the
```
takuro-sato:
```suggestion
// PostRawAttest retrieves a hardware attestation report signed by the
```

internal/httpginendpoints/httpginendpoints.go
```diff
@@ -0,0 +1,230 @@
...
+
+	rawReport, err := attest.RawAttest(inittimeDataBytes, runtimeDataBytes)
+	if err != nil {
+		c.JSON(http.StatusForbidden, gin.H{"error": err.Error()})
+	}
+
+	c.JSON(http.StatusOK, gin.H{"report": rawReport})
+}
+
+// postMAAAttest retrieves an attestation token issued by Microsoft Azure Attestation
```
takuro-sato:
```suggestion
// PostMAAAttest retrieves an attestation token issued by Microsoft Azure Attestation
```

internal/httpginendpoints/httpginendpoints.go
```diff
@@ -0,0 +1,230 @@
...
+
+	maaToken, err := certState.Attest(maa, runtimeDataBytes, *uvmInfo)
+	if err != nil {
+		c.JSON(http.StatusForbidden, gin.H{"error": err.Error()})
+	}
+
+	c.JSON(http.StatusOK, gin.H{"token": maaToken})
+}
+
+// postKeyRelease retrieves a secret previously imported to Azure Key Vault
```
takuro-sato:
```suggestion
// PostKeyRelease retrieves a secret previously imported to Azure Key Vault
```

internal/httpginendpoints/httpginendpoints.go
```diff
@@ -0,0 +1,230 @@
...
+	jwkJSONBytes, err := json.Marshal(jwKey)
+	if err != nil {
+		c.JSON(http.StatusForbidden, gin.H{"error": err.Error()})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{"key": string(jwkJSONBytes)})
+}
+
+func ApiMiddleware(certState *attest.CertState, identity *common.Identity, uvmInfo *common.UvmInformation) gin.HandlerFunc {
```
takuro-sato: How about using name like `RegisterGlobalStates` or something like that to indicate what it does?

takuro-sato (approved):

Issue #29, opened 2023-05-03
`pkg/skr` test is failing becuase of expired certificate


takuro-sato:
Maybe it's better to create the certificate on the fly rather than to use test data.

Used go version:
```log
root@SandboxHost-638186287547569439:~/confidential-sidecar-containers# go version
go version go1.16.15 linux/amd64
```

Failing test:

```log
root@SandboxHost-638186287547569439:~/confidential-sidecar-containers# go test -v ./...
?       github.com/Microsoft/confidential-sidecar-containers/cmd/azmount        [no test files]
Temporary directory: /tmp/aztemp101577859
=== RUN   Test_GetBytes_Supported
--- PASS: Test_GetBytes_Supported (0.00s)
=== RUN   Test_GetBytes_Unsupported
--- PASS: Test_GetBytes_Unsupported (0.00s)
=== RUN   Test_GetBlock_TestBounds
--- PASS: Test_GetBlock_TestBounds (0.00s)
PASS
ok      github.com/Microsoft/confidential-sidecar-containers/cmd/azmount/filemanager    (cached)
?       github.com/Microsoft/confidential-sidecar-containers/cmd/remotefs       [no test files]
?       github.com/Microsoft/confidential-sidecar-containers/cmd/skr    [no test files]
=== RUN   Test_SNP
=== RUN   Test_SNP/SerializeReport_Success
=== RUN   Test_SNP/DeserializeReport_Success
=== RUN   Test_SNP/DeserializeReport_InvalidSize
--- PASS: Test_SNP (0.00s)
    --- PASS: Test_SNP/SerializeReport_Success (0.00s)
    --- PASS: Test_SNP/DeserializeReport_Success (0.00s)
    --- PASS: Test_SNP/DeserializeReport_InvalidSize (0.00s)
=== RUN   Test_CertFetcher
=== RUN   Test_CertFetcher/CertFetcher_Success
=== RUN   Test_CertFetcher/CertFetcher_Invalid_PlatformVersion
=== RUN   Test_CertFetcher/CertFetcher_Invalid_ChipID
=== RUN   Test_CertFetcher/CertFetcher_Invalid_TEEType
=== RUN   Test_CertFetcher/CertFetcher_Invalid_EndpointType
--- PASS: Test_CertFetcher (0.69s)
    --- PASS: Test_CertFetcher/CertFetcher_Success (0.44s)
[...]

PR #37, opened 2023-05-15
Create test certs for each run to avoid expiration

takuro-sato:
Fix https://github.com/microsoft/confidential-sidecar-containers/issues/29


pkg/skr/skr_test.go
```diff
@@ -129,6 +133,93 @@ func Test_Keywrap(t *testing.T) {
...
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			CommonName: "Test Intermediate CA",
+		},
+		NotBefore: dateNow,
+		NotAfter:  dateTenYearsLater,
+
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+		MaxPathLen:            0,
```
takuro-sato, line 181: Functions under test look general x509 related functions rather than something very specific to SKR.
So I used default values for most of `x509.Certificate`'s fields so that test code looks simple.

pkg/skr/skr_test.go
```diff
@@ -129,6 +133,93 @@ func Test_Keywrap(t *testing.T) {
 	}
 }
 
+// Generates cert chain consists of root cert, intermediate cert, and leaf cert
+func generateTestCerts(dnsName string) ([]string, error) {
```
hgarvison, line 137: Should the same be done for pkg/attest/attest_test.go the Test_CertFetcher and Test_MAA functions?

BryceDFisher (approved):

pkg/skr/skr_test.go
takuro-sato: That's a very good point.
For Test_CertFetcher  we might want to stop using hard-coded cert for `expectedContent`'s value instead we can somehow validate the cert instead.
For Test_MAA, we can't generate certs that MAA accepts but we can fetch actual certs from a platform cert server (This cert will last until 2029 though).

If we want to address them now, I'll do so in this PR. Otherwise, I'll raise issue to run skipped tests first.

hgarvison: Since this PR is about creating test certs to avoid expiration and the fix for Test_CertFetcher and Test_MAA won't involve creating test certs, I think we can create a separate PR for those functions.

takuro-sato: Raised an issue https://github.com/microsoft/confidential-sidecar-containers/issues/38

KenGordon (approved):
I am happy with this.

PR #44, opened 2023-06-15
support for encrypted writeable filesystems 

kapilvgit:
This PR adds support for mounting encrypted writeable filesystems into confidential containers. The feature is exposed through an additional ```read_only``` attribute that can be specified in the filesystem configuration. This is true by default. Setting it to false allows containers to write to the mounted filesystem. Writes are cached locally and pushed to remote storage only if a block in the cache needs to be evicted, or the application calls fsync. 

The main changes are
- Additional logic in fuse to intercept writes
- Logic in the cache to write blocks back to storage on eviction.


examples/encfs/aci-arm-template2.json
```diff
@@ -33,10 +33,10 @@
                             "environmentVariables": [
                                 {
                                     "name": "EncfsSideCarArgs",
-                                    "value": "<base64-std-encodes-string-of-sidecar-args>"
+                                    "value": "ew0KICAgImF6dXJlX2ZpbGVzeXN0ZW1zIjogWw0KICAgICAgew0KICAgICAgICAgIm1vdW50X3BvaW50IjogIi9tbnQvcmVtb3RlL3NoYXJlIiwNCiAgICAgICAgICJhenVyZV91cmwiOiAiaHR0cHM6Ly9oZ2Fydmlzb25zdG9yYWdlLmJsb2IuY29yZS53aW5kb3dzLm5ldC9wcml2YXRlLWNvbnRhaW5lci9lbmNmcy5pbWciLA0KICAgICAgICAgImF6dXJlX3VybF9wcml2YXRlIjogdHJ1ZSwNCiAgICAgICAgICJyZWFkX29ubHkiOiBmYWxzZSwNCiAgICAgICAgICJrZXkiOiB7DQogICAgICAgICAgICAia2lkIjogImVuY2ZzLWRlbW8ta2V5LWhnYXJ2aXNvbiIsDQogICAgICAgICAgICAiYXV0aG9yaXR5Ijogew0KICAgICAgICAgICAgICAgImVuZHBvaW50IjogInNoYXJlZGV1czIuZXVzMi50ZXN0LmF0dGVzdC5henVyZS5uZXQiDQogICAgICAgICAgICB9LA0KICAgICAgICAgICAgImFrdiI6IHsNCiAgICAgICAgICAgICAgICJlbmRwb2ludCI6ICJhY2NtaHNtLm1hbmFnZWRoc20uYXp1cmUubmV0Ig0KICAgICAgICAgICAgfQ0KICAgICAgICAgfQ0KICAgICAgfQ0KICAgXQ0KfQ=="
```
KenGordon: This looks like a mistaken commit as it contains your actual args, experimental container location etc.

examples/encfs/encfs-sidecar-args.json
```diff
@@ -4,6 +4,7 @@
          "mount_point": "/mnt/remote/share",
          "azure_url": "<azure-blob-storage-url>",
          "azure_url_private": true,
+         "read_only": true,
```
KenGordon: Surely it is better to have a flag named after the new feature than the default that has to be added and explicitly switched off?

If the PR is about adding "read/write" I'd add a read_write flag rather than a !read_only flag.

cmd/azmount/filemanager/cache.go
```diff
@@ -72,6 +98,29 @@ func GetBlockSize() int64 {
 	return fm.blockSize
 }
 
+func IsReadOnly() bool {
+	return fm.readOnly
+}
+
+// Utility function to check if block is in the cache and download it if not
+func DownloadBlock(blockIndex int64) ([]byte, error) {
```
KenGordon: This function is not well named. Sometimes it download a block, sometimes not.

examples/encfs/aci-arm-template2.json
hgarvison: Good catch. Thanks.

cmd/azmount/filemanager/cache.go
hgarvison: Refactored into two functions with more appropriate names

examples/encfs/encfs-sidecar-args.json
hgarvison: Fixed

PR #47, opened 2023-07-07
Fix paths pointing to clients in SKR sidecar

DomAyre:
Previously paths for clients in the example arm template and README were /tests/*_client.sh but the dockerfile moves those scripts to /tests/skr/*_client.sh.

Fix the paths


DomAyre:
To test this, I copied the updated example arm template, populated it with my own MAA and HSM info, and deployed it with an allow_all policy.

Then I shelled into each container and checked the logs to ensure they now work as expected

KenGordon (approved):
Good.

PR #48, opened 2023-07-10
Create workflows for pushing SKR and EncFS images

DomAyre:
Adds workflows for each sidecar to deploy an image on PRs and merges to main.

Uses the existing build and push scripts to put images with the naming scheme {registry}/{sidecar}:{branch_name}.

These images can then be pulled and tested


.github/workflows/push_encfs_image.yml
```diff
@@ -0,0 +1,60 @@
+name: Push Encrypted FS Image
+
+on:
```
hgarvison, line 3: You could do "on: [push, pull-request]" and then list the branches and paths, since the paths and branches are the same. It'll make the files shorter and aid in readability.

.github/workflows/push_skr_image.yml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+
+  push-skr-image: 
+    name: Push SKR Image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        
+      - name: Log in to Azure Container Registry
```
hgarvison, line 33: There is a github action for this: https://github.com/marketplace/actions/azure-container-registry-login

.github/workflows/push_encfs_image.yml
DomAyre: Great suggestion I wasn't aware that was possible!

Addressed in https://github.com/microsoft/confidential-sidecar-containers/pull/48/commits/f33f49f1593755d3f4cfc67cdb03a8747c784360

.github/workflows/push_skr_image.yml
DomAyre: Good spot, I did use it in an earlier version but I got az login errors in a later step. But I think that was actually due to the push script doing a whole Azure CLI login not just the registry one which I fixed by making that optional so I should be able to bring this back

Done in https://github.com/microsoft/confidential-sidecar-containers/pull/48/commits/365a6d0946a93fba8f5a284c3b8a095f76d71e41

.github/workflows/push_encfs_image.yml
DomAyre: On second glance I'm not sure this is actually supported? I get a failure in the workflow which prevents it running (https://github.com/microsoft/confidential-sidecar-containers/actions/runs/5518492850) and when I do a GitHub wide code search for an example which specifies two triggers which share branches and paths (https://github.com/search?q=%5Bpull_request%2C+push%5D+path%3A%2F%5E.github%5C%2Fworkflows%5C%2F%2F&type=code) I can't find any examples which do that.

So I'll revert for now, then if I've done it wrong we can make another change to fix later

KenGordon: @DomAyre I am happy this whole PR is small enough in scope not to hurt the repo in general. Please merge when you are happy with the operation.

KenGordon (approved):

.github/workflows/push_skr_image.yml
DomAyre: It turns out that login through this action isn't recognised by the run step after all so I'll revert this change too. I did experiment with using the docker build and push action together with the login action which I use for the examples repo, but I think here it's good to have coverage of the build/push.sh scripts

DomAyre:
The latest image generated by this change was used to run the key_release example 
https://github.com/microsoft/confidential-aci-examples/actions/runs/5520428801 which passed

.github/workflows/push_encfs_image.yml
hgarvison: I think it needs to have the "on: [push, pull_request]" on the same line. From your commit, it looked like you had the list on the next and indented. https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#:~:text=forks%20the%20repository%3A-,on%3A%20%5Bpush%2C%20fork%5D,-If%20you%20specify

PR #49, opened 2023-07-11
Simplify tag derivation for image pushing

DomAyre:
The original PR anticipated a difference in where the branch name would be in the GitHub context depending on the event trigger, this turned out not to be needed so it's been removed.

Also fixed credential passing and branch fetching for the cleanup workflow as this couldn't be tested by triggering the workflow until the workflow existed on main


hgarvison (approved):

KenGordon (approved):
OK, safe and outside of the main sidecar path.

PR #50, opened 2023-07-12
Allow credentials access from fork PRs

DomAyre:
Use the pull request target event instead of pull request which allows repository secrets access under the basis that it runs against the target branch so malicious PRs cannot leak the secret without already merging something to main. We then checkout the PR head which makes it equivalent to using pull_request only with access to the secrets.

The safety of this approach is based on the fact that this repository (at least at the time of writing) requires approval from someone with write access for a PR from an outside fork to run the actions. This means the this is only safe under the following conditions:
- We don't change the setting allowing anyone to run actions on their fork PRs
- All admins of this repo are careful which PRs they approve to run actions

### DIFF
--- a/.github/workflows/push_encfs_image.yml
+++ b/.github/workflows/push_encfs_image.yml
@@ -5,6 +5,13 @@ on:
   pull_request_target:
     branches:
       - main
+    paths:
+      - .github/workflows/push_encfs_image.yml
+      - cmd/azmount/**
+      - cmd/remotefs/**
+      - docker/encfs/**
+      - tools/get-snp-report/bin/get-snp-report
+      - tools/get-snp-report/bin/get-fake-snp-report
   push:
     branches:
       - main
--- a/.github/workflows/push_encfs_image.yml
+++ b/.github/workflows/push_encfs_image.yml
@@ -5,13 +5,6 @@ on:
   pull_request_target:
     branches:
       - main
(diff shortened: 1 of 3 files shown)


DomAyre:
Unfortunately since pull_request_target runs against the target branch so we won't see it run until the PR is merged

KenGordon (approved):

.github/workflows/push_encfs_image.yml
```diff
@@ -41,11 +43,16 @@ jobs:
...
+          if [ ${{ github.event_name }} == "push" ]; then
+            branch_name=main
+          else  
+            branch_name=$(echo ${{ github.head_ref }})
+          fi
           docker/encfs/push.sh \
             ${{ secrets.REGISTRY_NAME }} \
             ${{ secrets.REGISTRY_DOMAIN }} \
-            encfs:${{ github.head_ref }} \
+            skr:$branch_name \
```
hgarvison: should be "encfs:$branch_name"

DomAyre: Thanks for spotting this, fixed in https://github.com/microsoft/confidential-sidecar-containers/pull/50/commits/f2bda00d77a4835b16ca1ba5a22c97353953363a

hgarvison (approved):

